Issue #24, opened 2020-01-08
support pygeos


geospatial-jeff:
I've found that [`pygeos`](https://github.com/pygeos/pygeos) can significantly speed up the creation of mosaicJSON by optimizing the [spatial operation to find assets which intersect each parent tile](https://github.com/developmentseed/cogeo-mosaic/blob/master/cogeo_mosaic/utils.py#L313-L325), especially when generating larger mosaics.

Pygeos should be [interoperable with shapely](https://github.com/pygeos/pygeos/blob/master/pygeos/io.py#L187-L223) on the next release (see https://github.com/Toblerity/Shapely/issues/782 for context), at which point I think it would make sense to add here.

vincentsarago:
Thanks @geospatial-jeff, I'll have a look or if you want to open a PR I'll be happy to help

geospatial-jeff:
I can do the PR, I'll start working on it this weekend!

vincentsarago:
closed in #25 #26 thanks a lot @geospatial-jeff

PR #25, opened 2020-01-13
Pygeos

geospatial-jeff:
- Replaces shapely operations with pygeos when creating/updating mosaicJSON.

The shapely dependency could probably be removed but it conveniently packages GEOS with its `.whl`.

Resolves https://github.com/developmentseed/cogeo-mosaic/issues/24

### DIFF
--- a/cogeo_mosaic/utils.py
+++ b/cogeo_mosaic/utils.py
@@ -18,9 +18,8 @@ import click
 import requests
 
 import numpy
-
+from pygeos import intersects, intersection, polygons, area
 import mercantile
-from shapely.geometry import shape, box
 from supermercado import burntiles
 
 import rasterio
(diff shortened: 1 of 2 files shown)


vincentsarago:
🎉 thanks for the PR @geospatial-jeff I'll try to review/merge/deploy today

vincentsarago:
> The shapely dependency could probably be removed but it conveniently packages GEOS with its .whl.

@geospatial-jeff so there is now wheels for pygeos! For lambda this is ok because right now we use a lambda layer with geos lib installed. This will be a problem for CI testing but I guess we can fix that.

We could add a note in the readme mentioning that `geos-config` is required.

vincentsarago:
as mentioned ☝️ tests are going to failed because we can't install pygeos without lib-geos being installed in the testing docker image.

I'll fix this

Issue #37, opened 2020-03-26
Refactor MosaicJSON parsing into base classes


kylebarron:

vincentsarago:
Thanks for the issue @kylebarron. 

This came up after https://github.com/developmentseed/cogeo-mosaic-tiler/issues/3. 

Basically the idea is to add a DynamoDB Backend to store the mosaicJSON information (using quadkey as primary key. 

To enable DynamoDB we first need to refactor the way to load and read the mosaicJSON in https://github.com/developmentseed/cogeo-mosaic/blob/master/cogeo_mosaic/utils.py#L397-L488. 

To do:
- [ ] create ABC Base class
  Must have these methods
    - metadata
    - tile (find asset for a tile)
    - point (find asset for a coordinate)
    - create 
    - update
- [ ] add File Backend (no schema)
- [ ] add http Backend (http, https)
- [ ] add S3 Backend (s3://)
- [ ] add DynamoDB backend (dynamodb://? ) 

Will be great to write those class to be used like
```python
with DynamoDBBackedn(url) as mosaic:
     return mosaic.tile(qk)
```

We need to think how the new backend play with https://github.com/developmentseed/cogeo-mosaic/blob/master/cogeo_mosaic/utils.py#L397-L421 and how this will be transparent for the tiler! 

cc @kylebarron @geospatial-jeff

kylebarron:
> Will be great to write those class to be used like
> 
> ```python
> with DynamoDBBackedn(url) as mosaic:
>      return mosaic.tile(qk)
> ```

I love it. I think that just means defining a couple context manager methods.

> We need to think how the new backend play with

I would think just call `BackendClass.tile()` instead of `fetch_and_find_assets`,  `BackendClass.point()` instead of `fetch_and_find_assets_point`, etc.

> how this will be transparent for the tiler

Somewhere just have a `switch`:
```py
backendClass
if scheme == 's3':
	BackendClass = S3Backend
elif scheme in ['http', 'https']:
	BackendClass = HttpBackend
elif scheme = 'dynamodb':
	BackendClass = DynamoDBBackend
else:
	BackendClass = FileBackend

with BackendClass(*args, **kwargs) as mosaic:
	mosaic.tile(x, y, z)
	mosaic.point(lng, lat)
```
?

kylebarron:
[...]

PR #38, opened 2020-03-26
Backend Plugins

kylebarron:
MosaicJSON backend plugins. Ref #37

Todo in future PRs:

- [ ] read assets at arbitrary `bbox`, `z`


cogeo_mosaic/backend/base.py
```diff
@@ -0,0 +1,30 @@
...
+
+import abc
+import contextlib
+
+
+class BaseBackend(metaclass=contextlib.AbstractContextManager):
+    @abc.abstractmethod
+    def __enter__(self, *args, **kwargs):
+        """Load resource"""
+
```
vincentsarago: definitely not an expert but I think we still need to set an `__init__`class where we can set some metadata 
we could say that each backend has an `path` or `url` for example (or all the mosaic metadata)

https://github.com/rafelafrance/traiter_vertnet/blob/fc95fbdff9ebc7ab569d24f7b3e5ee644d2c6ad6/vertnet/writers/base_writer.py#L7-L21

Also I think by default __enter__ should return self (like in ☝️ ) and we should also have a __exit__ (it won't do anything exept maybe for closing some db connection)

cogeo_mosaic/backend/s3.py
```diff
@@ -0,0 +1,51 @@
...
+
+from cogeo_mosaic.backend.base import BaseBackend
+from cogeo_mosaic.backend.utils import get_assets_from_json
+from cogeo_mosaic.utils import _decompress_gz
+
+
+class S3Backend(BaseBackend):
+    """S3 Backend Adapter"""
+
+    def __init__(self, bucket: str, key: str):
```
vincentsarago: can we assume we are passing a full s3 path here ? and maybe an optional client ?

cogeo_mosaic/backend/base.py
kylebarron: Yes, both `__init__` and `__enter__` are required.

I think that by inheriting from `contextlib.AbstractContextManager` it creates a default `def __enter__: return self`: https://docs.python.org/3/library/contextlib.html#contextlib.AbstractContextManager

cogeo_mosaic/backend/dynamo.py
```diff
@@ -0,0 +1,81 @@
...
+        min_zoom = self.mosaic_def["minzoom"]
+        quadkey_zoom = self.mosaic_def.get("quadkey_zoom", min_zoom)  # 0.0.2
+        tile = mercantile.tile(lng, lat, quadkey_zoom)
+        return self.get_assets(tile.x, tile.y, tile.z)
+
+    @functools.lru_cache(maxsize=512)
+    def fetch_mosaic_definition(self) -> Dict:
+        """Get Mosaic definition info."""
+        # NOTE(kylebarron): may also want to convert Decimal to int/float here
+        return fetch_dynamodb("-1")
```
vincentsarago: ok so `-1` is a dummy `Key` where we store the a json object with the mosaicJSON metadata right ? 

dumb question, does the Key has to be a number ?

kylebarron: Yes, `-1` is the dummy `Key` for storing non-`tiles` metadata.

The Key has to be the same data type throughout. Since quadkeys are strings, all `quadkey` values need to be strings, hence `'-1'`

vincentsarago: thanks

cogeo_mosaic/backend/s3.py
kylebarron: > can we assume we are passing a full s3 path here

I think yes... If we assume the url is of type
```
s3://<bucket>/<key>
```
then we always have the full path to the bucket.

If we maintain the current `/mosaicid/{endpoint}` then we can add the bucket prefix there before passing to `S3Backend`

> and maybe an optional client

I added `client` instantiation to `__init__`, so you'll be able to do
```py
with S3Backend(bucket, key) as mosaic:
	s3client = mosaic.client
```

vincentsarago: well but idealy when we do
```
with S3Backend(bucket, key) as mosaic:
```
the __enter__ method will fetch the mosaic using a client 

that why I think doing 
```
with S3Backend(s3path, client=myclient) as mosaic:
```
would be better

vincentsarago: > If we maintain the current /mosaicid/{endpoint} then we can add the bucket prefix there before passing to S3Backend` 

The mosaicid based endpoint should be considered as `custom` and maybe removed ? ir maybe we add an environment variable telling which is our fall back method (s3 or dynamo)...

kylebarron: `__init__` is called on `__enter__` I believe. Putting logic in `__init__` and having `__enter__` return just `self` looks to be common:
https://stackoverflow.com/a/34749985

vincentsarago: maybe I'm not exlaining mysefl right. 

the idea is to define the s3 client outside the backend 

```
from boto3.session import Session as boto3_session
session = boto3_session()
client = session.client("s3")

with S3Backend(s3path, client=myclient) as mosaic:
    assets = mosaic.tile(1,1,1)
```

cogeo_mosaic/backend/base.py
```diff
@@ -0,0 +1,31 @@
+"""cogeo_mosaic.backend.base: base Backend class."""
+
+import abc
+import contextlib
+from typing import Dict
+
+
+class BaseBackend(metaclass=contextlib.AbstractContextManager):
```
kylebarron: I'm not sure if this is correct. https://github.com/rafelafrance/traiter_vertnet/blob/fc95fbdff9ebc7ab569d24f7b3e5ee644d2c6ad6/vertnet/writers/base_writer.py#L4-L7
does it slightly differently.

I got this code from https://pymotw.com/3/abc/

cogeo_mosaic/backend/s3.py
kylebarron: Ok, yeah, I added an optional `client` arg here:
https://github.com/developmentseed/cogeo-mosaic/pull/38/files#diff-998203b2b4af94384b3c4a81c56c57f5R19

Maybe I should do that for `dynamodb` too... though I'm using `boto3.resource` for it, not `boto3.Session.client`... not sure what the difference is

cogeo_mosaic/backend/base.py
vincentsarago: I also do it differently https://github.com/vincentsarago/rio-tiler-experiment/blob/master/cog_tiler/base.py#L14 (mine is definitely worse). Not sure what is the best

cogeo_mosaic/backend/s3.py
vincentsarago: cool, 
they are mostly the same, just different ways to call them. I problem boto3_session because I think you could pass more options to it, but that's maybe not true

kylebarron: My `dynamodb` code only seems to work with `boto3.resource`, not `boto3.Session().client()` so I'll leave it for now

vincentsarago: cool

cogeo_mosaic/backend/file.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class FileBackend(BaseBackend):
+    """Local File Backend Adapter"""
+
+    def __init__(self, path: str):
+        self.mosaic_def = self.fetch_mosaic_definition(path)
+
+    def metadata(self) -> Dict:
+        return self.mosaic_def
```
vincentsarago: I think it would be good to keep consistency between each backend. e.g dynamoDB will return only the metadata while here we will return the full mosaicJSON.

kylebarron: This is true, but I'm not sure how to get around that. It's not possible to return the entire `tiles` key for DynamoDB, so either `tiles` is _sometimes_ returned, or _never_ returned?

cogeo_mosaic/backend/dynamo.py
```diff
@@ -0,0 +1,81 @@
...
+import itertools
+import os
+from typing import Dict, Tuple
+
+import mercantile
+
+from cogeo_mosaic.backend.base import BaseBackend
+
+
+class DynamoDBBackend(BaseBackend):
```
vincentsarago: Question, is it possible to return the list of `Keys`?

cogeo_mosaic/backend/file.py
kylebarron: Also I think some of these default things should go in the base class, like the default
```py
    def metadata(self) -> Dict:
        return self.mosaic_def
```

cogeo_mosaic/backend/dynamo.py
kylebarron: That would mean querying the whole table I think? So I would assume no without a performance penalty

cogeo_mosaic/backend/file.py
vincentsarago: yeah tiles (or quadkeys) could be optional

cogeo_mosaic/backend/dynamo.py
vincentsarago: cool cool

kylebarron: Maybe it's possible to do a faster scan (https://stackoverflow.com/a/48734264) to get the key names, but it seems not ideal.

kylebarron: So it is true that you don't know if a quadkey exists until you query DynamoDB, but that query is <10ms, so not a big deal

vincentsarago: that's totally fine 
I was just asking because we use `quakeys` to create geojson somewhere. but again that's optional

cogeo_mosaic/backend/file.py
vincentsarago: > Also I think some of these default things should go in the base class, like the default

yeah but again `metadata` should be for metadata not the full mosaicJSON IMO

kylebarron: As it is it's optional; it exists in all classes other than DynamoDB. That doesn't conform to the MosaicJSON spec, so maybe create an empty `tiles` array?

vincentsarago: > That doesn't conform to the MosaicJSON spec, so maybe create an empty tiles array

Well it's not true that it doesn't conform to the spec, yes it's not a JSON docs but it must have the same info (metadata + quadkey base index)

> so maybe create an empty tiles array?
no just raise an exception 
```
raise NotAvailable("{} is not available for DynamoDB backend")
```

cogeo_mosaic/backend/base.py
```diff
@@ -0,0 +1,31 @@
...
+    @abc.abstractmethod
+    def __init__(self, *args, **kwargs):
+        """Load resource"""
```
vincentsarago: what do you think about adding more properties in the __init__ describing the metadata of the mosaicjson? 

```
def __init__(self, *args, **kwargs):
    self.minzoom: int
    self.maxzoom: int
    self.quadkey_zoom: Optional[int]
    self.bounds: Tuple[float, float, float, float]
    self.center: Optional[Tuple[float, float, int]]
    sefl.version: Optional[str]
    self.name: Optional[str]
    self.mosaicjson_version: str
```

cogeo_mosaic/backend/base.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def metadata(self) -> Dict:
+        """Retrieve MosaicJSON metadata."""
+        return self.mosaic_def
```
vincentsarago: as mentioned before, this makes thing not really clear if metadata is returning metadata about the mosaic or the full mosaicjson document

cogeo_mosaic/backend/dynamo.py
```diff
@@ -0,0 +1,88 @@
...
+    """DynamoDB Backend Adapter"""
+
+    def __init__(
+        self, mosaicid: str, region: str = os.getenv("AWS_REGION", "us-east-1")
+    ):
+        self.client = boto3.resource("dynamodb", region_name=region)
+        self.table = self.client.Table(mosaicid)
+        self.mosaic_def = self.fetch_mosaic_definition()
```
vincentsarago: maybe fecth_mosaic_definion could just feed properties define here https://github.com/developmentseed/cogeo-mosaic/pull/38/files#r398594144

cogeo_mosaic/backend/dynamo.py
```diff
@@ -0,0 +1,88 @@
...
+            mercator_tiles = [mercator_tile]
+            for i in range(depth):
+                mercator_tiles = sum(
+                    [mercantile.children(t) for t in mercator_tiles], []
+                )
+
+            mercator_tiles = list(filter(lambda t: t.z == quadkey_zoom, mercator_tiles))
+            quadkey = [mercantile.quadkey(*tile) for tile in mercator_tiles]
+        else:
+            quadkey = [mercantile.quadkey(*mercator_tile)]
```
vincentsarago: Can we make this block (get_parent, get_child...) an `.utils` function ?

cogeo_mosaic/backend/base.py
kylebarron: I don't really have a preference. Currently they're wrapped in a `dict`, unwrapping them is fine too

cogeo_mosaic/backend/base.py
kylebarron: I'm not sure the best way to solve that? The difference is whether `tiles` has the full array or not?

cogeo_mosaic/backend/dynamo.py
kylebarron: Yeah. I think that's good. I get a little confused by the itertools recursion. 

The function in `utils` can take an `accessor` argument that defaults to `.get()` for a `dict` but could be replaced by `.fetch_dynamodb()`

cogeo_mosaic/backend/file.py
kylebarron: I'm confused. You want to raise an error instead of returning _any_ metadata? 

I'd be in favor of returning all metadata except for `tiles`, and just have a note in the docs that a function using `Backend.metadata` can't depend on the `tiles` key existing.

cogeo_mosaic/backend/base.py
kylebarron: As it is, essentially the only keys used are `quadkey_zoom`, `minzoom`, and `maxzoom`, right? It might be cleaner to keep them in a `dict`?

cogeo_mosaic/backend/base.py
kylebarron: I think maybe it would be best to never return `tiles`, or at least default to not...
So
```py
def metadata(self, tiles=False) -> Dict:
	"""Retrieve Mosaic metadata

	Args:
	- tiles: Include `tiles` key in response

	Returns: MosaicJSON as dict, not including the `tiles` key unless specified
	"""
	if tiles:
		if 'tiles' not in self.mosaic_def.keys():
			raise ValueError...
		return self.mosaic_def

	return {k: v for k, v in self.mosaic_def.items() if k != 'tiles'}
```

This way

1. It can be consistent across all backends, even for potential new backends  in the future where it's also hard to get a list of all keys
2. Minimizes default return payload size, since as we've seen including `tiles` can make the full mosaicjson huge.

cogeo_mosaic/backend/file.py
kylebarron: I'm resolving this here; lets continue discussion on the comment about `metadata` class method

cogeo_mosaic/backend/base.py
vincentsarago: > I think maybe it would be best to never return tiles

That's what I meant

cogeo_mosaic/backend/base.py
vincentsarago: Yes and no, because it could make it cleaner when we only need to access values 
so instead of doing 
```
return self.mosaic_def["minzoom"] 
```
we do 
```
return self.minzoom
```

cogeo_mosaic/backend/dynamo.py
vincentsarago: yeah we don't need to over simplified!

cogeo_mosaic/backend/dynamo.py
kylebarron: I think it's easiest to return a `dict` from `fetch_mosaic_definition`... Otherwise you'd be returning 8 values that you need to unpack into `self` right? Like
```py
self.minzoom, self.maxzoom, self.quadkey_zoom ... = self.fetch_mosaic_definition()
```
Is that what you mean?

cogeo_mosaic/backend/file.py
vincentsarago: yeah that's what I mean, 
metadata should return the mosaic metadata (zoom, bounds ...) but not the quadkey list (tiles) 

this (tiles) should be covered by another method/properties which should raise an error in dynamoDB

cogeo_mosaic/backend/base.py
kylebarron: It wasn't working with `metaclass` so I switched to the use in my first link.

cogeo_mosaic/backend/dynamo.py
vincentsarago: if you define the values in init in the base class you can then do 
```
class Base(ABC):
def __init__(self, *args, **kwargs):
    self.minzoom: int
    self.maxzoom: int
    self.quadkey_zoom: Optional[int]
    self.bounds: Tuple[float, float, float, float]
    self.center: Optional[Tuple[float, float, int]]
    sefl.version: Optional[str]
    self.name: Optional[str]
    self.mosaicjson_version: str
```
```
class backend(Base):
    def __init__(self, url, ...):
        super(backend, self).__init__()
        self._fetch_mosaic_def(url)
```

 the `_fetch_mosaic_def` will do:

```
def _fetch_mosaic_def(self, url):
    mosaic = fetch(...)
    self.minzoom = mosaic["minzoom"]
....

```

vincentsarago: this is maybe totally not the best way to do it (just thinking how I would have done it :D )

cogeo_mosaic/backend/base.py
kylebarron: Ok, I misunderstood, but I think now it's fixed as you wanted?

vincentsarago: yeah I still have strong opinion about `metadata` should return metadata of the mosaic definition not the tiles (even with an option) but we can revisit this later

kylebarron: I suppose you're right... What are the use cases for calling `metadata`? 
Essentially only the `tile()` and `point()` methods would need to access `tiles`, right?

cogeo_mosaic/backend/dynamo.py
kylebarron: I refactored the quadkey finding into a separate function.

cogeo_mosaic/backend/base.py
kylebarron: Ok I updated to never return `tiles`

cogeo_mosaic/backend/dynamo.py
kylebarron: Currently the only one of those we actually use is `quadkey_zoom` (derived from it and `minzoom`). So I defined only `quadkey_zoom` in `__init__` and if we need more values in the future we can add them then. Sound good?

cogeo_mosaic/backend/base.py
kylebarron: As I mentioned in another comment:
> Currently the only one of those we actually use is quadkey_zoom (derived from it and minzoom). So I defined only quadkey_zoom in __init__ and if we need more values in the future we can add them then. Sound good?

cogeo_mosaic/utils.py
```diff
@@ -394,95 +386,95 @@ def update_mosaic(
...
-        itertools.chain.from_iterable(
-            [
-                fetch_and_find_assets(asset, x, y, z)
-                if os.path.splitext(asset)[1] in [".json", ".gz"]
-                else [asset]
-                for asset in assets
-            ]
-        )
-    )
+# def get_mosaic_content(url: str) -> Dict:
```
kylebarron: @vincentsarago 
All this commented out code is now redundant, but removing it would presumably be a breaking change. Is a breaking change ok?

kylebarron:
Currently we load the MosaicJSON immediately on `__init__`... That won't always work, especially for the upload/create/update methods, so it should probably be in a `try/except`? Or have an option to `__init__` to not attempt to load the mosaicjson

vincentsarago:
oh yeah I didn't think about the `create` method, maybe this can happen outside the backends? 
For the update, we will still need to be able to fetch the mosaic.

cogeo_mosaic/utils.py
vincentsarago: breaking change is totally fine ;-)

kylebarron:
> oh yeah I didn't think about the `create` method, maybe this can happen outside the backends?

I currently have a dummy `create` method that just calls `create_mosaic` and then a separate `upload` method for each backend. I'm not sure what `update` does exactly... It adds a single asset to a mosaic? So you'd need to grab the mosaic, add the asset and put it back?

cogeo_mosaic/backend/base.py
kylebarron: It looks like `__enter__` and `__exit__` aren't created by default, so I added them to the base class.

kylebarron:
The base class looks to work now with a Context Manager. Aka this works correctly now:
```py
with S3Backend(bucket=bucket, key=key) as mosaic:
    print(mosaic.quadkey_zoom) # 8

table_name = '7610d6d77fca346802fb21b89668cb12ef3162a31eb71734a8aaf5de'
region = 'us-west-2'
with DynamoDBBackend(table_name=table_name, region=region) as mosaic:
    print(mosaic.quadkey_zoom) # 12
```

### Load mosaicJSON on `__init__`?

It's generally most convenient to load the mosaic JSON when the class is instantiated, because then you can access `self.mosaic_def` and `self.quadkey_zoom` in later methods.

The issue here is that the `create`/`upload` and `update` methods don't necessarily have an existing mosaicJSON. So loading a MosaicJSON on `__init__` when it won't be used anyways wastes time and might error anyways.

Right now I added a `load_mosaic` option, though I don't really like it. There could be a `mode` option, similar to that in the `open` built-in, so `mode='r'` reads the MosaicJSON, but `mode='w'` doesn't because it expects to write only. Whereas `mode='rw'` or `mode='a'` for append could be used with the `update` method.

### Required methods?

Should `update` and `upload/create` be required methods? Specifically I'm thinking of the `HTTPBackend`... how would an arbitrary `upload` method be implemented in that? A `POST` to the given url?

cogeo_mosaic/backend/base.py
```diff
@@ -0,0 +1,70 @@
...
+        -------
+        MosaicJSON as dict without `tiles` key.
+        """
+        if self.mosaic_def is None:
+            return {}
+
+        return {k: v for k, v in self.mosaic_def.items() if k != "tiles"}
+
+    @abc.abstractmethod
+    def tile(self, x: int, y: int, z: int):
```
geospatial-jeff, line 39: I think using input param of bbox is more flexible (all XYZ tiles have defined bbox but not all bbox have a specific XYZ tile).  This gives more flexibility in terms of caching strategies -- if you know you want to do a lot of tile requests within a specific area you can cache the asset lookup for that entire area.

cogeo_mosaic/backend/http.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def point(self, lng: float, lat: float) -> Tuple[str]:
+        """Retrieve assets for point."""
+        tile = mercantile.tile(lng, lat, self.quadkey_zoom)
+        return get_assets_from_json(
+            self.mosaic_def["tiles"], self.quadkey_zoom, tile.x, tile.y, tile.z
+        )
+
+    @functools.lru_cache(maxsize=512)
+    def fetch_mosaic_definition(self, url: str) -> Dict:
```
geospatial-jeff, line 33: Should be an abstract method in the base class.  Maybe rename to `read` as well to be consistent with CRUD

cogeo_mosaic/backend/base.py
```diff
@@ -0,0 +1,70 @@
...
+
+        return None
+
+    @abc.abstractmethod
+    def upload(self, mosaic: Dict):
+        """Upload new MosaicJSON to backend."""
+
+    # Commented to allow child classes to not implement this
+    # @abc.abstractmethod
+    def update(self, mosaic: Dict):
```
geospatial-jeff, line 69: Give mosaicJSON is a well defined spec, I think we should add some stricter validation than standard python types.  I'm a big fan of `pydantic` which will throw validation errors if the mosaicJSON doesn't match the defined model.  This will make it easier to add new backends because the mosaicJSON definition is explicitly defined in the code and prevent exceptions (ex `KeyError`) from happening downstream.

```python
from pydantic import BaseModel

class MosaicJSON(BaseModel):
    """https://github.com/developmentseed/mosaicjson-spec/tree/master/0.0.2"""

    minzoom: int
    maxzoom: int
    tiles: Dict[str, List[str]]
    bounds: List[NumType] = [-180, -90, 180, 90]
    name: Optional[str] = "null"
    description: Optional[str] = "null"
    attribution: Optional[str] = "null"
    center: Optional[List[NumType]] = "null"
    version: Optional[str] = "1.0.0"
    quadkey_zoom: Optional[int] = 0
    mosaicjson: Optional[str] = "0.0.2"
```

cogeo_mosaic/backend/base.py
```diff
@@ -0,0 +1,70 @@
...
+from cogeo_mosaic.backend.utils import get_hash
+from cogeo_mosaic.utils import create_mosaic
+
+
+class BaseBackend(AbstractContextManager):
+    mosaic_def: Optional[Dict]
+
+    @abc.abstractmethod
+    def __init__(self, *args, **kwargs):
+        """Connect to backend"""
```
geospatial-jeff, line 17: Connecting to any resources required to access the backend should happen within `__enter__` and cleaned up within `__exit__`

cogeo_mosaic/backend/base.py
```diff
@@ -0,0 +1,70 @@
...
+
+from cogeo_mosaic import version as mosaic_version
+from cogeo_mosaic.backend.utils import get_hash
+from cogeo_mosaic.utils import create_mosaic
+
+
+class BaseBackend(AbstractContextManager):
+    mosaic_def: Optional[Dict]
+
+    @abc.abstractmethod
```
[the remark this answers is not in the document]
kylebarron: Up to @vincentsarago what the minimum Python required is, but since 3.6's end of life is the end of 2021, I assume we'll stick with 3.6.

cogeo_mosaic/backend/base.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self, *args, **kwargs):
+        """Connect to backend"""
+
+    def __enter__(self):
```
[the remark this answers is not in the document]
kylebarron: Well my thought was that since `__enter__` always calls `__init__`, it's easiest to just have the default base `__enter__` and `__exit__` be non-abstract methods, so that the child classes don't actually need to overwrite `__enter__`, just `__init__`.

cogeo_mosaic/backend/base.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @abc.abstractmethod
+    def tile(self, x: int, y: int, z: int):
+        """Retrieve assets for tile."""
+
+    @abc.abstractmethod
+    def point(self, lng: float, lat: float):
+        """Retrieve assets for point."""
+
+    def create(self, *args, **kwargs):
```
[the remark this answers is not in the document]
kylebarron: > That being said, the process to create a mosaicJSON is pretty similar once you have a list of assets

That's why I have `create` as a non-`abstractmethod` and `upload` as the `abstractmethod`. `create` is the same for all backends once you have the list of URLs. Right now `create` calls `self.upload` automatically, but I could change that to `return`. In my current terminology, I think of `create` as the method to take the urls and create a MosaicJSON, and `upload` as taking a pre-formed MosaicJSON and putting it on the backend.

> I don't think that the base backend would specify how a child class creates a mosaicjson since that is specific to the particular backend.

My thought process is that no matter what backend you'll end up uploading to, an intermediate `dict` representation of a MosaicJSON is ideal. This way each backend's `upload` method can always assume it takes in a fully-formed MosaicJSON.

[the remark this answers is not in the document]
kylebarron: > Personally I prefer the pattern of attaching helper functions which are used by children classes as methods of the base class (self._create_mosaic) but I'm not opposed to leaving it as is.

Isn't that basically what I'm doing?

cogeo_mosaic/backend/base.py
```diff
@@ -0,0 +1,70 @@
...
+    @property
+    def mosaicid(self) -> str:
+        return get_hash(body=self.mosaic_def, version=mosaic_version)
+
+    @property
+    def quadkey_zoom(self) -> Optional[int]:
+        if self.mosaic_def:
+            return self.mosaic_def.get("quadkey_zoom", self.mosaic_def["minzoom"])
+
+        return None
```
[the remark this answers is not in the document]
kylebarron: Well the issue is that `mosaic_def` doesn't always exist. Specifically, for the `upload`/`create` methods, `mosaic_def` doesn't exist during `__init__`. You can see that if a `mosaic_def` exists, I define `quadkey_zoom` by first checking for the `dict`'s `quadkey_zoom`, and next taking its `minzoom`.

cogeo_mosaic/backend/base.py
kylebarron: Ooh does this mean once you define your code as listed you can just pass
```
def update(self, mosaic: MosaicJSON):
```
? That's really cool. Although I don't think `mypy` is currently actually checking types, though @vincentsarago can correct me if I'm wrong.

cogeo_mosaic/backend/http.py
kylebarron: Good point

cogeo_mosaic/backend/base.py
kylebarron: 1. Can we be certain that there won't be float precision issues where we ask for one xyz tile and actually retrieve 9? 

2). I feel that it really depends on the internals. If the internals are all in terms of mercator tiles, then it's going to be simplest to just pass down the xyz index. If we pass down a `bbox` that doesn't align with a mercator tile, what does it do? Find all the mercator tiles that overlap with the bbox? Return a custom image with the given pixel density for that `bbox`?

cogeo_mosaic/backend/base.py
kylebarron: I guess I just thought it was simpler to do it inside `__init__` ... Is there ever a time when `__init__` is called but not `__enter__`?

cogeo_mosaic/backend/base.py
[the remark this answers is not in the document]
vincentsarago: I don't use python 3.6 but that might be a really strong decision to not support 3.6 🤔 
let's keep it for now, I don't think it adds too much code

cogeo_mosaic/backend/base.py
vincentsarago: I love this idea. 
here the version of the model I had on my side.

```python
class MosaicJSON(BaseModel):
    """
    MosaicJSON model.

    Based on https://github.com/developmentseed/mosaicjson-spec

    """

    mosaicjson: str = "0.0.2"
    name: Optional[str]
    description: Optional[str]
    version: str = "1.0.0"
    attribution: Optional[str]
    minzoom: int = Field(0, ge=0, le=30)
    maxzoom: int = Field(30, ge=0, le=30)
    quadkey_zoom: Optional[int]
    bounds: List[float] = [-180, -90, 180, 90]
    tiler: Dict[str, List[str]]
```

cogeo_mosaic/backend/base.py
vincentsarago: > I feel that it really depends on the internals. If the internals are all in terms of mercator tiles, then it's going to be simplest to just pass down the xyz index. If we pass down a bbox that doesn't align with a mercator tile, what does it do? Find all the mercator tiles that overlap with the bbox? Return a custom image with the given pixel density for that bbox?

☝️ totally agree with this. 

I think we could add an other method to retrieve asset from `bbox, z`

cogeo_mosaic/backend/base.py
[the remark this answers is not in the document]
geospatial-jeff: > In my current terminology, I think of create as the method to take the urls and create a MosaicJSON

But not every backend will use the logic defined in `cogeo_mosaic.utils.create_mosaic`.  For example if you have a PostGIS database of STAC Items, you don't need to read the footprint of each URL because you already have that information from querying the database.

> but I could change that to return

I like this idea, its best if each method does one thing.

cogeo_mosaic/backend/base.py
geospatial-jeff: Yeah `__init__` is called whenever an object is created.  `__enter__` is only called when the object is entered using a `with` statement.

cogeo_mosaic/backend/base.py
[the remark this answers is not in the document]
geospatial-jeff: > so that the child classes don't actually need to overwrite __enter__

If this is the case we should remove `__enter__` and `__exit__` from the base class because they are already defined in `abc.AbstractContextManager`.

cogeo_mosaic/backend/base.py
kylebarron: I guess I wasn't sure that I wanted to force always using a context manager. Like when you put necessary logic in `__init__`, it means that 

1. Someone can choose not to use a context manager if they don't want to
2. There's never an occasion when a class variable is expected to be defined, but is only defined in `__enter__` instead of `__init__` so code unexpectedly breaks

cogeo_mosaic/backend/base.py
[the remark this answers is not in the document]
geospatial-jeff: Makes sense, I also misread your code you are already defaulting to minzoom

cogeo_mosaic/backend/base.py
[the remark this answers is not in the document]
kylebarron: That's what I thought but then when I actually ran the code it gave me errors unless I specifically added `__enter__` and `__exit__` as methods.

vincentsarago:
@kylebarron  I still believe that adding `create/update` within the backend classes adds a certain complexity.

What would you think if we just add `utility` function within the backend code like
```
from cogeo_mosaic.backend.s3 import upload
or
from cogeo_mosaic.backend.dynamo import upload
```

Same for the `update` method. Right now the update method needs the list of quadkeys to recalulate the bbox of the mosaic, but we could fix this by buring the actual bbox of the mosaic and adding the new quadkeys, so it could also work for DynamoDB

```
    updated_qk = []

    dataset_geoms = polygons([feat["geometry"]["coordinates"][0] for feat in results])
    for idx, r in enumerate(results):
        tiles = burntiles.burn([r], quadkey_zoom)
        tiles = ["{2}-{0}-{1}".format(*tile.tolist()) for tile in tiles]

        dataset = [{"path": r["properties"]["path"], "geometry": dataset_geoms[idx]}]
        for parent in tiles:
            z, x, y = list(map(int, parent.split("-")))
            parent = mercantile.Tile(x=x, y=y, z=z)
            quad = mercantile.quadkey(*parent)
            tile_geometry = polygons(
                mercantile.feature(parent)["geometry"]["coordinates"][0]
            )

            fdataset = dataset
            if minimum_tile_cover is not None:
                fdataset = _filter_and_sort(
                    tile_geometry, fdataset, minimum_cover=minimum_tile_cover
                )

            if len(fdataset):
                dst_quad = mosaic_def["tiles"].get(quad, [])
                for f in fdataset:
                    dst_quad.append(f["path"])

                mosaic_def["tiles"][quad] = dst_quad  # THis should either update the DB or update the mosaicjson doc.
                updated_qk.append(quad)

    tiles = mercantile.tiles(*mosaic_def["bounds"], quadkey_zoom) + updated_qk
    bounds = tiles_to_bounds(tiles)

    mosaic_def["bounds"] = bounds
    mosaic_def["center"] = [
        (bounds[0] + bounds[2]) / 2,
        (bounds[1] + bounds[3]) / 2,
        mosaic_def["minzoom"],
    ]
```

cogeo_mosaic/backend/base.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @abc.abstractmethod
+    def upload(self, mosaic: Dict):
+        """Upload new MosaicJSON to backend."""
+
+    # Commented to allow child classes to not implement this
```
geospatial-jeff, line 67: Better to force child classes to do:
```
def update(...):
    raise NotImplementedError
```

cogeo_mosaic/backend/dynamo.py
```diff
@@ -0,0 +1,135 @@
...
+    ):
+        self.client = boto3.resource("dynamodb", region_name=region)
+        self.table_name = table_name
+        self.mosaic_def = None
+
+        if table_name:
+            self.table = self.client.Table(table_name)
+
+            if load_mosaic:
+                self.mosaic_def = self.fetch_mosaic_definition()
```
vincentsarago, line 35: I'm not sure about the `load_mosaic` option, 

I think the idea is to load the mosaic when we are in a `read` mode ? 

maybe something like 
```
    def __init__(
         self,
         table_name: str,
         region: str = os.getenv("AWS_REGION", "us-east-1"),
         mosaic_def: Optional[Dict] = None,
     ):
    # Write
     if mosaic_def:
        # Here we know we should upload the mosaic_definition
        # Note: add a check if the table already exist ? 
     
     # Read
     else:
          self.table = self.client.Table(table_name)
          self.mosaic_def = self.fetch_mosaic_definition()
```

cogeo_mosaic/backend/base.py
geospatial-jeff: I don't think it really matters if we use the backends as context managers or not, as long as we are consistent (so the call pattern across backends is the same) and the usage is clear from the base class.  Inheriting from `AbstractContextManager` makes me think that backends are supposed to be used with context.

cogeo_mosaic/backend/dynamo.py
kylebarron: Yeah, that's possible... I think the issue is what you do for the `create/` endpoint, when you haven't created a Mosaic yet. _Or_ you just require that a mosaic be created outside of the backends, which might be simpler

cogeo_mosaic/backend/base.py
kylebarron: I added this to a "future todo" list in the original comment, and we can take it up in a future PR.

cogeo_mosaic/backend/http.py
kylebarron: Made an abstract method and renamed to `read_mosaic`. I feel that `read` would be a little too confusing with `tile`, since this is only reading the mosaic, not any actual data assets.

cogeo_mosaic/backend/base.py
kylebarron: Do we require that `tiles` exist? For example, loading the Mosaic definition from DynamoDB doesn't include the `tiles` key. So to be consistent, we could either allow `tiles` to be `Optional[Dict[str, List[str]]]`, or when loading the mosaic definition from dynamodb add an empty `dict`

kylebarron: For now, I just set `tiles` to be an empty `dict` when loading from DynamoDB

cogeo_mosaic/backend/base.py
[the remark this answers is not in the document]
kylebarron: I think it's probably that `abc.AbstractContextManager` defines `__enter__` and `__exit__` as _abstract_ methods. Since we really don't have any setup/teardown, I think defining `__enter__` and `__exit__` as non-abstract methods in the base class are fine, and then we really don't need to subclass from `abc.AbstractContextManager` anyways

cogeo_mosaic/backend/base.py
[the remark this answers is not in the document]
kylebarron: I removed this entirely, as Mosaic creation should happen entirely outside of the backends. Specifically I now require a Mosaic to exist in `__init__`. I think that if one can't be fetched or isn't provided with the right fields an error will be produced, either from failing to fetch or through pydantic checking

cogeo_mosaic/backend/base.py
kylebarron: I also have `self.mosaic_def` stored internally as an instance of the pydantic class, instead of a plain `dict`

cogeo_mosaic/backend/base.py
kylebarron: Ok, I guess I don't have super strong feelings, but let me know if you think I should change something

cogeo_mosaic/backend/dynamo.py
kylebarron: I think I updated to an acceptable solution, where `mosaic_def` is required to be fetched from the url/bucket+key/dynamo table/file name, **or provided in `__init__`**. If neither is the case, an error is raised I think

kylebarron:
Ok I think this PR is ready to be looked at again. I still haven't implemented an `update` method, partially because I don't understand exactly what needs to happen there. Could that come in the next PR, since I would think `update` is the least used of all methods?

I defined the pydantic MosaicJSON model in a separate file `model.py`. The Mosaic definition is always held internally as a pydantic class, so it should raise an error if a user passes an invalid dict to the class.

I think requiring `mosaic_def` to either be provided in `__init__` as an arg or be fetched is a great solution. It means that we can always assume in other methods that `self.mosaic_def` exists. And no weird edge cases should be needed for `upload`/`update`. (Well, with `update` I guess you do need to handle _two_ mosaicJSONs, the existing and the new. So I'd propose documenting that the one in `__init__` should always be an existing mosaicjson, and the update method can handle a new mosaic as an arg.)

> I still believe that adding `create/update` within the backend classes adds a certain complexity.

Check the current code and see if you like it. I think the complexity is way down.

> Same for the `update` method. Right now the update method needs the list of quadkeys to recalulate the bbox of the mosaic, but we could fix this by buring the actual bbox of the mosaic and adding the new quadkeys, so it could also work for DynamoDB

I'm a bit confused by that code snippet. What are you trying to do?

kylebarron:
Should we have a custom exception for when a mosaic definition either can't be loaded from the provided info or when the provided mosaic definition dict fails pydantic's test?

cogeo_mosaic/backend/base.py
vincentsarago: @kylebarron the the `tiles` I think it's before to raise an exception, the `tiles` is theoretically set but just not accessible with dynamoDB

cogeo_mosaic/backends/base.py
```diff
@@ -0,0 +1,61 @@
...
+    def point(self, lng: float, lat: float) -> Tuple[str]:
+        """Retrieve assets for point."""
+
+    @abc.abstractmethod
+    def read_mosaic(self, *args, **kwargs) -> MosaicJSON:
+        """Fetch mosaic definition"""
+
+    @property
+    def mosaicid(self) -> str:
+        return get_hash(body=dict(self.mosaic_def), version=mosaic_version)
```
vincentsarago, line 49: I don't think we need to add the `version` here, the mosaicid should be uniq per mosaic, by adding the version, each time the cogeo-mosaic module will be updated the mosaic id will change.

cogeo_mosaic/backends/base.py
```diff
@@ -0,0 +1,61 @@
...
+    @abc.abstractmethod
+    def tile(self, x: int, y: int, z: int) -> Tuple[str]:
+        """Retrieve assets for tile."""
+
+    @abc.abstractmethod
+    def point(self, lng: float, lat: float) -> Tuple[str]:
+        """Retrieve assets for point."""
+
+    @abc.abstractmethod
+    def read_mosaic(self, *args, **kwargs) -> MosaicJSON:
```
vincentsarago, line 44: not sure if `read` will make more sense here 
cc @geospatial-jeff @kylebarron

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -0,0 +1,140 @@
...
+
+import boto3
+import mercantile
+
+from cogeo_mosaic.backends.base import BaseBackend
+from cogeo_mosaic.backends.utils import find_quadkeys
+from cogeo_mosaic.model import MosaicJSON
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
```
vincentsarago, line 17: 🤔 I wonder if the logger shouldn't be set in the backend itself. right now I'm not sure `logger.warn` will show up later because we set it here with `INFO`

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -0,0 +1,140 @@
...
+
+class DynamoDBBackend(BaseBackend):
+    """DynamoDB Backend Adapter"""
+
+    def __init__(
+        self,
+        table_name: str,
+        mosaic_def: Optional[Union[MosaicJSON, Dict]] = None,
+        region: str = os.getenv("AWS_REGION", "us-east-1"),
+        client: Optional[Any] = None,
```
vincentsarago, line 28: is `client` really `Any` ?

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def upload(self):
+        self._create_table()
+        items = self._create_items()
+        self._upload_items(items)
+
+    def update(self):
+        raise NotImplementedError
+
+    def _create_table(self, billing_mode: str = "PAY_PER_REQUEST"):
```
vincentsarago, line 55: maybe we could use `enum` for billing_mode, so we make sure we only pass good values

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -0,0 +1,140 @@
...
+        for key in ["bounds", "center"]:
+            if meta.get(key):
+                meta[key] = list(map(float, meta[key]))
+
+        # Create pydantic class
+        # For now, a tiles key must exist
+        meta["tiles"] = {}
+        return MosaicJSON(**meta)
+
+    def get_assets(self, x: int, y: int, z: int) -> Tuple[str]:
```
vincentsarago, line 126: Not sure but maybe we could add LRU cache here too ?

cogeo_mosaic/backends/base.py
```diff
@@ -0,0 +1,61 @@
...
+    @property
+    def mosaicid(self) -> str:
+        return get_hash(body=dict(self.mosaic_def), version=mosaic_version)
+
+    @property
+    def quadkey_zoom(self) -> Optional[int]:
+        return self.mosaic_def.quadkey_zoom or self.mosaic_def.minzoom
+
+    @abc.abstractmethod
+    def upload(self):
```
vincentsarago, line 56: if we use `read` as proposed before, maybe we can use `write` here?

cogeo_mosaic/backends/file.py
```diff
@@ -0,0 +1,56 @@
...
+    def point(self, lng: float, lat: float) -> Tuple[str]:
+        """Retrieve assets for point."""
+        tile = mercantile.tile(lng, lat, self.quadkey_zoom)
+        return get_assets_from_json(
+            self.mosaic_def.tiles, self.quadkey_zoom, tile.x, tile.y, tile.z
+        )
+
+    def upload(self):
+        with open(self.path, "wb") as f:
+            f.write(_compress_gz_json(dict(self.mosaic_def)))
```
vincentsarago, line 42: compression could be an option ?

cogeo_mosaic/backends/auto.py
```diff
@@ -0,0 +1,38 @@
...
+
+from urllib.parse import urlparse
+
+from cogeo_mosaic.backends.dynamodb import DynamoDBBackend
+from cogeo_mosaic.backends.file import FileBackend
+from cogeo_mosaic.backends.http import HttpBackend
+from cogeo_mosaic.backends.s3 import S3Backend
+
+
+def auto_backend(url: str, **kwargs):
```
vincentsarago, line 15: I'm not 💯 sure about the `name` auto_backend but can't come up with a better name. If we assume this will be the most used method maybe something like `MosaicReader`? 

```python
with MosaicReader(url) as mosaic:
    asset = mosaic.tile(x, y, z) 
```

cogeo_mosaic/backends/base.py
kylebarron: Yes I agree. I only add the version because you did so before. In fact the hash computed here still won't match one computed from the command line, because it's wrapped in another dict `{'body': ...`

cogeo_mosaic/backends/base.py
kylebarron: I thought that read would be too confusing because you're not actually reading any data. A `read` method on a backend class would generally make me think I'm reading data (assets), not just the metadata

cogeo_mosaic/backends/dynamodb.py
kylebarron: I always just use print myself 😅 so I'll leave this to you

cogeo_mosaic/backends/dynamodb.py
kylebarron: I couldn't find the right class for it, because it's not the same class as the S3 one. Instead it's a `boto3.resource()`

cogeo_mosaic/backends/dynamodb.py
kylebarron: I have to look up how to do this, and this option currently isn't exposed to the user at all

cogeo_mosaic/backends/dynamodb.py
kylebarron: I guess it doesn't hurt

cogeo_mosaic/backends/auto.py
kylebarron: Well clearly I'm bad at naming 😅. @geospatial-jeff ?

cogeo_mosaic/backends/file.py
kylebarron: Good point. Also when compression is used should a `.gz` suffix be added to the filename? Both here and with S3?

cogeo_mosaic/backends/base.py
vincentsarago: so just do 
```
get_hash(**dict(self.mosaic_def)) 
```
?

cogeo_mosaic/backends/base.py
vincentsarago: I think `read` is just saying the backend is accessing its data which is, in this case, the mosaic definition.

cogeo_mosaic/backends/base.py
kylebarron: Yes, that should be equivalent.

cogeo_mosaic/backends/file.py
vincentsarago: I think the `.gz` is already in the filename -> so we apply compression

but I'm not 💯 sure this is the best way 🤔

kylebarron: Maybe
```
def upload(self, gzip=None):
```
if `gzip` is `True`, apply compression, if `False`, don't, and if `None` apply compression if `url`/`key`/`path` ends in `.gz`

vincentsarago: sounds good

kylebarron:
I'll try to update in response to a few comments tonight, but if things otherwise look good, I'd suggest merging and iterating (small changes, docs, tests) in future PRs, because this PR is incredibly long and bloated already (129 comments)

cogeo_mosaic/backend/base.py
kylebarron: I don't know what you meant?

vincentsarago: I mean `tiles` shouldn't return empty because in theory `tiles` is present in the mosaicjson document, it's just that dynamoDB backend is not able to return the whole `tiles` array. That's why I think this should raise an exception in this backend

cogeo_mosaic/backends/auto.py
kylebarron: Renamed to `MosaicBackend`

cogeo_mosaic/backends/file.py
kylebarron: Updated to be smarter with compression as described above.

kylebarron:
Ok I think I dealt with all the comments. The CI is failing because some of the imports in other files, e.g. the `cli` module have invalid imports now.

I'd suggest merging and then doing final clean up in future PRs. Then we can split future PRs into a docs PR, a tests PR, etc.

vincentsarago:
@kylebarron I don't think we should merge this to master. I'm going to make a `v3` branch which we could use for the next PR

PR #43, opened 2020-04-15
Use R-tree for intersection tests when creating mosaic

### LINKED ISSUE #42: Use rtree when creating mosaic
Currently, the algorithm used when generating the mosaic is relatively slow: for every tile, it checks for intersection between that tile and _every asset bounds_. 
https://github.com/developmentseed/cogeo-mosaic/blob/bfe427c102129a8627750889d7b50341c2aa4c11/cogeo_mosaic/utils.py#L301-L304

I see you're using `pygeos` here, which is indeed faster than Shapely, but it still has to compute every pairwise intersection.

That can be made faster by using a tree.

Pygeos even includes a tree! https://pygeos.readthedocs.io/en/latest/strtree.html

And even cooler, it has a [`query_bulk`](https://pygeos.readthedocs.io/en/latest/strtree.html#pygeos.strtree.STRtree.query_bulk) method for vectorized querying of the tree! So you can pass all the tile geometries at once, and then have very few to double check intersections of.

cc @geospatial-jeff

kylebarron:
Closes #42 , cc @geospatial-jeff 

This uses an R-tree from pygeos for testing intersections. 

As part of the speedup, the `tree.query` can both find candidate results _and_ check set theoretic methods. So I don't have to do two passes, once to get results, and a second time to double check intersections. These timings are updated for the new, faster code.

With 22,081 Landsat assets (predownloaded from Sat API) and 535 mercator tiles, it takes about 250ms to make the mosaic. This doesn't include fetching asset metadata or finding the tiles. (With existing code it's 600ms.)

With 220,000 Landsat assets, it's 9s with existing code and 2.2s with the tree.

Finding the tiles using `burntiles.burn` is kinda slow (500ms), but it finds the tiles that actually intersect with at least one asset, not just taking the maximum bounds and finding all tiles within those bounds. That means it created only ~500 tiles, instead of the 2000 tiles that would have been found by naively retrieving all tiles inside the max bounds.

Using `pygeos.total_bounds` instead of `burntiles.find_extrema` made it 200ms faster (with 22,000 assets).


kylebarron:
Well I finally fixed all the imports, then ~sometime soon I can fix why the tests are failing

PR #44, opened 2020-04-16
Backend Plugins (#38)

vincentsarago:
Continue work from #38


kylebarron:
Since `.metadata` doesn't return the `tiles` key of the MosaicJSON, I don't think the `/geojson` endpoint provided by `awspds-mosaic` is possible with the current backends?

vincentsarago:
> Since .metadata doesn't return the tiles key of the MosaicJSON, I don't think the /geojson endpoint provided by awspds-mosaic is possible with the current backends?

@kylebarron well this is still possible by using the `self.mosaic_def`

kylebarron:
Oh right, it's just that the `tiles` key might not exist.

cogeo_mosaic/backends/file.py
```diff
@@ -0,0 +1,66 @@
...
+        """Retrieve assets for point."""
+        tile = mercantile.tile(lng, lat, self.quadkey_zoom)
+        return get_assets_from_json(
+            self.mosaic_def.tiles, self.quadkey_zoom, tile.x, tile.y, tile.z
+        )
+
+    def write(self, gzip=None):
+        """Write mosaicjson document to a file."""
+        body = dict(self.mosaic_def)
+        if gzip or (gzip is None and self.path.endswith(".gz")):
```
vincentsarago, line 45: @kylebarron @geospatial-jeff I think I mis understood this one.

if an user wants to create a mosaic file, basically I was expecting we could do:

```python
with MosaicBackend(None, mosaic_def=mosaic_content) as mosaic:
    mosaic.write("mosaic.json")
```
This won't work because self.path is set to `None` and the function doesn't allow `"mosaic.json"` as input. 

for now the correct way to do is: 

```python
with MosaicBackend("mosaic.json", mosaic_def=mosaic_content) as mosaic:
    mosaic.write()
```

any though on what's the best way to go forward ?

cogeo_mosaic/backends/file.py
```diff
@@ -0,0 +1,66 @@
...
+            body = _compress_gz_json(body)
+        else:
+            body = json.dumps(body).encode("utf-8")
+
+        with open(self.path, "wb") as f:
```
vincentsarago, line 50: Should we allow to save mosaic in non-binary format ?

cogeo_mosaic/backends/file.py
kylebarron: > ```python
> with MosaicBackend("mosaic.json", mosaic_def=mosaic_content) as mosaic:
>     mosaic.write()
> ```

I think this is fine, because it mimics `open()`

cogeo_mosaic/backends/file.py
kylebarron: Well a "text format" is binary encoded as UTF-8, so I don't see a difference.

cogeo_mosaic/backends/file.py
vincentsarago: ok so we will need to add a test to make sure `path` is set to something

vincentsarago: > ok so we will need to add a test to make sure path is set to something

well this might not be needed, what if someone just want to create a mosaic in memory without reading it (ok this is an edge case), but we might request a path will it won't be used 
🤔

kylebarron: > what if someone just want to create a mosaic in memory without reading it

I think that's a perfect example for a custom backend. If someone wants that, they can have a "backend" that writes to a `BytesIO` or `StringIO` buffer.

cogeo_mosaic/backends/base.py
```diff
@@ -0,0 +1,66 @@
...
+    @abc.abstractmethod
+    def tile(self, x: int, y: int, z: int) -> List[str]:
+        """Retrieve assets for tile."""
+
+    @abc.abstractmethod
+    def point(self, lng: float, lat: float) -> List[str]:
+        """Retrieve assets for point."""
+
+    @abc.abstractmethod
+    def read(self) -> MosaicJSON:
```
vincentsarago, line 47: @kylebarron @geospatial-jeff Should `read` be a private method?
It's not really needed by the user and will/should only be used by the class itself

vincentsarago:
Alright, I'm just waiting for feedback on the `read` method and on https://github.com/developmentseed/cogeo-mosaic/pull/54. but when done, I'll merge this branch into master so we can merge all the other PR. 

I've started some documentation but we'll need more. 

cogeo_mosaic/backends/base.py
kylebarron: I agree that it would be uncommon for a user to use it, but I don't necessarily think it should be private.

vincentsarago: so in fact I learn something, `_` is not really **private** but more like semiprivate, while `__` is **private** 

> one _ simply to indicate to other programmers that the attribute or method is intended to be private

ref: https://stackoverflow.com/a/1301369

I do think `read` is should be at least `semiprivate`

kylebarron: I don't follow your conclusion... Both `_` and `__` are private, but `__` has special meanings within a class.

I also don't agree that it will never be called by the user. What if someone wanted to implement a custom merging between an existing mosaic and one on a backend? They'd do:
```py
with MosaicBackend(local_mosaic, url=...) as mosaic:
	remote_mosaic = mosaic.read()
```
Because if you call `mosaic.mosaic_def`, it gives you the `local_mosaic` you passed in.

vincentsarago: > I don't follow your conclusion... Both _ and __ are private, but __ has special meanings within a class.

well it seems that `_` just serve as an indication, while `__` makes the method fully privat to a class

vincentsarago:
```
with MosaicBackend(local_mosaic, url=...) as mosaic:
	remote_mosaic = mosaic.read()
	# logic to combine mosaics
```

gonna have fun with this one.

I guess this could work better with a `mosaic.merge(another_mosaic)`

```python
with CustomMosaicBackend(local_mosaic) as mosaic:
    with MosaicBackend("https://....") as mosaic_2:
	    mosaic.merge(mosaic_2)
```

kylebarron: > I guess this could work better with a `mosaic.merge(another_mosaic)`

But then we're getting more and more specific. How should the mosaics be merged? That's a big question with several different possible solutions.

While I'm not opposed to adding such a `merge` method, I think that the user should be able to call `read` to implement custom merging. And by defining `read` as a public method, we're saying that's behavior that's specifically enabled to the user.

kylebarron: The point of the backends is to provide a common base for further work to be done on top of. `read` is a basic crud method, and I think it would be wrong to make it private, so as to tell users not to use it. It's fine to write in the docs that most people won't need to use it, but that doesn't imply it should be private.

vincentsarago: I'm not saying we should add a merge method... you bring that up ;-). I just showed what could be a cleaner way to do instead of passing 2 mosaics (local_mosaic and url=) in the MosaicBackend ...

kylebarron: > I'm not saying we should add a merge method... you bring that up

No, I specifically argued for the opposite... that instead of us creating a `merge` method, we should make public the methods that would enable a user to create a merge method.

I don't see what's wrong with calling it `read` but saying in the docstring "a user normally won't call this directly". I think that's quite common in Python packages.

vincentsarago: > No, I specifically argued for the opposite... that instead of us creating a merge method, we should make public the methods that would enable a user to create a merge method.

In my example, the `merge` was part of the `CustomMosaicBackend`, I won't implement this ;-)

vincentsarago: at this point, I still think that if a user do need to call `read` it should be in a really advanced case so it shouldn't matter for the user if it's `read` or `_read`. But from simple user perspective, I believe having `read` as a public method could be kinda misleading because in 99.9% of the time people won't need the `read` method.

PR #45, opened 2020-04-16
edits on Kyle master PR

vincentsarago:
This PR does: 
- move `cogeo_mosaic.backends.auto.MosaicBackend` to `cogeo_mosaic.backends.MosaicBackend`
- add missing docstring (to be edited) 
- remove the `update` CLI (To be added back in future PR)
-  re-order module imports (good habits to put typing first)
- make `backend.metadata` a property
- remove logging in DynamoDBBackend and add progressbar (TBD)

cc @kylebarron


cogeo_mosaic/backends/file.py
```diff
@@ -38,23 +40,27 @@ def point(self, lng: float, lat: float) -> Tuple[str]:
         )
 
     def write(self, gzip=None):
+        """Write mosaicjson document to a file."""
         body = dict(self.mosaic_def)
         if gzip or (gzip is None and self.path.endswith(".gz")):
             body = _compress_gz_json(body)
+        else:
+            body = json.dumps(body).encode("utf-8")
```
vincentsarago, line 48: we always write a binary file

cogeo_mosaic/backends/file.py
```diff
@@ -38,23 +40,27 @@ def point(self, lng: float, lat: float) -> Tuple[str]:
...
 
     def update(self):
+        """Update the mosaicjson document."""
         raise NotImplementedError
 
     @functools.lru_cache(maxsize=512)
-    def read(self, path: str) -> MosaicJSON:
-        """Get Mosaic definition info."""
-        with open(path, "rb") as f:
+    def read(self) -> MosaicJSON:
```
vincentsarago, line 58: we don't need `path` in the input because it is set in `self.path`

kylebarron:
> * re-order module imports (good habits to put typing first)

I always order my imports with [`isort`](https://github.com/timothycrosley/isort). I love it because it's deterministic, so I never have to give a thought towards import ordering.

cogeo_mosaic/backends/base.py
```diff
@@ -40,15 +44,17 @@ def point(self, lng: float, lat: float) -> Tuple[str]:
         """Retrieve assets for point."""
 
     @abc.abstractmethod
-    def read(self, *args, **kwargs) -> MosaicJSON:
+    def read(self) -> MosaicJSON:
```
kylebarron, line 47: At some point I got some linter errors when the base class's `read` had a different function signature than the inherited classes

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -66,20 +70,20 @@ def _create_table(self, billing_mode: str = "PAY_PER_REQUEST"):
...
-            logger.info("creating table")
 
             # If outside try/except block, could wait forever if unable to
             # create table
             self.table.wait_until_exists()
-        except boto3.client("dynamodb").exceptions.ResourceInUseException:
-            logger.warn("unable to create table, may already exist")
+        except self.client.exceptions.ResourceInUseException:
+            warnings.warn("Unable to create table, may already exist")
+            return
```
kylebarron, line 79: You don't really need this `return`, but I guess that's more a stylistic decision

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -124,18 +125,19 @@ def read(self) -> MosaicJSON:
         return MosaicJSON(**meta)
 
     @functools.lru_cache(maxsize=512)
-    def get_assets(self, x: int, y: int, z: int) -> Tuple[str]:
+    def get_assets(self, x: int, y: int, z: int) -> Sequence[str]:
```
kylebarron, line 128: Why `Sequence` instead of `Tuple`?

cogeo_mosaic/backends/base.py
```diff
@@ -1,14 +1,17 @@
 """cogeo_mosaic.backend.base: base Backend class."""
 
+from typing import Dict, Tuple
+
```
kylebarron, line 4: I think the default `isort` behavior is group imports into 

1. Standard lib
2. External packages
3. Internal modulels

Within each group, put all `import`s before all `from ... import`

And then alphabetize within each subgroup.

cogeo_mosaic/backends/file.py
```diff
@@ -1,33 +1,35 @@
...
 
 
 class FileBackend(BaseBackend):
     """Local File Backend Adapter"""
 
-    def __init__(
-        self, path: str, mosaic_def: Optional[Union[MosaicJSON, Dict]] = None,
-    ):
+    def __init__(self, path: str, mosaic_def: Optional[Union[MosaicJSON, Dict]] = None):
+        """Initialize FileBackend."""
```
kylebarron, line 23: Was this reformatted with black? It's surprising because I ran black on everything. Does black pick up the longer line-length here: https://github.com/developmentseed/cogeo-mosaic/blob/60749264bd8375c482ca5420ccefd39964edb3ce/tox.ini#L44
?

cogeo_mosaic/backends/file.py
kylebarron: Yeah I was wondering if that was an issue. I guess this is equivalent to switching between `open(..., 'w')` and `open(..., 'wb')`

cogeo_mosaic/backends/s3.py
```diff
@@ -24,69 +28,71 @@ def __init__(
...
+        """Initialize S3Backend."""
         self.client = client or boto3_session().client("s3")
         self.key = key
         self.bucket = bucket
 
         if mosaic_def is not None:
             self.mosaic_def = MosaicJSON(**dict(mosaic_def))
         else:
-            self.mosaic_def = self.read(bucket, key)
+            self.mosaic_def = self.read()
```
kylebarron, line 39: I just tend to pass the non-`self` methods inside `__init__` in case in the future the objects defined in `__init__` get moved around or something

cogeo_mosaic/backends/s3.py
```diff
@@ -24,69 +28,71 @@ def __init__(
...
+        """Write mosaicjson document to AWS S3."""
         body = dict(self.mosaic_def)
         if gzip or (gzip is None and self.key.endswith(".gz")):
             body = _compress_gz_json(body)
+        else:
+            body = json.dumps(body).encode("utf-8")
 
-        _aws_put_data(
-            self.key, self.bucket, body, client=self.client,
-        )
```
kylebarron, line 56: Again, I'm assuming black is picking up on a different line length that you have set

cogeo_mosaic/backends/utils.py
```diff
@@ -1,26 +1,31 @@
...
     Attributes
     ----------
-    mercator_tile: mercantile.Tile
-        Input tile to use when searching for quadkeys
-    quadkey_zoom: int
-        Zoom level
+        mercator_tile: mercantile.Tile
+            Input tile to use when searching for quadkeys
+        quadkey_zoom: int
+            Zoom level
```
kylebarron, line 23: I don't think the extra indentation is standard, within the package, e.g.:
https://github.com/developmentseed/cogeo-mosaic/blob/60749264bd8375c482ca5420ccefd39964edb3ce/cogeo_mosaic/utils.py#L87-L97

cogeo_mosaic/utils.py
```diff
@@ -135,8 +135,8 @@ def _get_point(asset: str, coordinates: Tuple[float, float]) -> dict:
             )
 
             if not (
-                (src_dst.bounds[0] < lng_srs[0] < src_dst.bounds[2]) and # noqa
-                (src_dst.bounds[1] < lat_srs[0] < src_dst.bounds[3])
+                (src_dst.bounds[0] < lng_srs[0] < src_dst.bounds[2])
+                and (src_dst.bounds[1] < lat_srs[0] < src_dst.bounds[3])  # noqa
```
kylebarron, line 139: I guess you might not need `# noqa` here, because of 
https://github.com/developmentseed/cogeo-mosaic/blob/60749264bd8375c482ca5420ccefd39964edb3ce/.pre-commit-config.yaml#L19-L22

Which my `flake8` wasn't picking up on

cogeo_mosaic/backends/base.py
vincentsarago: I think the linter errors were because of the `@lru_cache` wrapper

cogeo_mosaic/backends/dynamodb.py
vincentsarago: Because we are returning a `list`. I could put `List` but I think Sequence is usually better

cogeo_mosaic/backends/base.py
vincentsarago: yeah I will tend to agree with isort, but let's keep the ordering as similar as is the other module for now.

cogeo_mosaic/backends/file.py
vincentsarago: yes. we usually use 90

cogeo_mosaic/backends/utils.py
vincentsarago: yeah, to be honest I'm not sure, I found that adding this extra indentation enable pydoc preview within VSCode, So I'm trying to add this every time I have to touch a function. (e. g I updated most of rio-tiler code like this)

PR #46, opened 2020-04-21
Add custom accessor option to create mosaic function

kylebarron:
I was reimplementing mosaic creation with pygeos in my `awspds_mosaic` forks/PRs, and got annoyed with the duplication. 

This PR refactors the mosaic creation function, exposing an internal `create_mosaic_from_features` function that takes a list of GeoJSON Features. This internal function just computes the mosaic from parameters, and leaves the checking to the higher level function.

Crucially, this PR also allows **an argument to be a _custom accessor_,** a lambda function that a user can provide to define how to find the identifier within the GeoJSON. Because this lambda function can be user provided, it's general enough to be included in the base `cogeo-mosaic` package.

The default accessor is equivalent to the existing code:
```py
accessor = lambda feature: feature["path"]
```
This retrieves the `path` attribute and sets it as the asset's identifier in the quadkey.

But because it's completely general, the same code could be used for, e.g. landsat, by just passing:
```py
accessor = lambda feature: feature['properties']['landsat:product_id']
```

Since this is an "advanced feature", it doesn't need to be provided through the CLI, which might be a little confusing, but this simple change allows a third party to easily define any value in the dict, nested or not, as the identifier.

As another example, since the identifier might also be the `id` key of the GeoJSON Feature, a user could simply pass `lambda feature: feature['id']` to make that the identifier.

This is currently built on #43 , since I'm using both in my fork, but could be refactored onto the existing code, if you decided not to accept #43.


cogeo_mosaic/create.py
```diff
@@ -58,10 +53,10 @@ def create_mosaic(
     if not quiet:
         click.echo("Get files footprint", err=True)
 
-    results = get_footprints(dataset_list, max_threads=max_threads, quiet=quiet)
+    features = get_footprints(dataset_list, max_threads=max_threads, quiet=quiet)
 
     if minzoom is None:
-        minzoom = list(set([feat["properties"]["minzoom"] for feat in results]))
+        minzoom = {feat["properties"]["minzoom"] for feat in features}
```
vincentsarago, line 59: 💥

cogeo_mosaic/create.py
```diff
@@ -58,10 +53,10 @@ def create_mosaic(
     if not quiet:
         click.echo("Get files footprint", err=True)
 
-    results = get_footprints(dataset_list, max_threads=max_threads, quiet=quiet)
+    features = get_footprints(dataset_list, max_threads=max_threads, quiet=quiet)
```
vincentsarago, line 56: 👍

cogeo_mosaic/create.py
kylebarron: So clean!

cogeo_mosaic/create.py
```diff
@@ -12,44 +11,40 @@
 
 
 def create_mosaic(
-    dataset_list: Sequence[str],
```
kylebarron, line 15: Should this be changed back to `Sequence`? I think I might've missed that in the merge conflicts

vincentsarago: I'll fix that

cogeo_mosaic/create.py
```diff
@@ -123,10 +171,10 @@ def create_mosaic(
             continue
 
         intersections_geoms = [dataset_geoms[idx] for idx in intersections_idx]
-        intersections = [results[idx] for idx in intersections_idx]
+        intersections = [features[idx] for idx in intersections_idx]
 
         dataset = [
-            {"path": f["properties"]["path"], "geometry": geom}
+            {"identifier": accessor(f), "geometry": geom}
```
kylebarron, line 177: I realized the `accessor` should actually get called here currently. In #59 I refactor this a little so that you don't need to create a new `list` of `dict`s.

cogeo_mosaic/create.py
kylebarron: I'll just push it back

kylebarron: Back to `Sequence`

cogeo_mosaic/create.py
```diff
@@ -70,28 +65,81 @@ def create_mosaic(
...
+        features,
+        minzoom=minzoom,
+        maxzoom=maxzoom,
+        version=version,
+        quiet=quiet,
+        **kwargs,
+    )
+
+
+def create_mosaic_from_features(
```
kylebarron, line 90: So any suggestion on what this should be called? I don't think it should be private. It's just that the above function is a higher-level overview for this.

cogeo_mosaic/create.py
kylebarron: What I was saying before is that now this function is essentially just a wrapper to call `get_footprints`, do some error checking, and pass `features` on.

cogeo_mosaic/create.py
```diff
@@ -15,41 +14,37 @@ def create_mosaic(
...
+            mosaicJSON definition version
+        quiet: bool, optional (default: True)
+            Mask processing steps.
 
     Returns
     -------
-    mosaic_definition : dict
-        Mosaic definition.
+        mosaic_definition : dict
+            Mosaic definition.
```
vincentsarago, line 47: I though I reverted the additional padding @kylebarron

cogeo_mosaic/create.py
vincentsarago: `from_features` sounds good to me (because it's in `mosaic.py`)

kylebarron: Well currently this is `create.py`

cogeo_mosaic/create.py
kylebarron: Maybe that was a merge conflict

PR #54, opened 2020-04-22
forward gzip option

vincentsarago:


vincentsarago:
@kylebarron what do you think about enabling `gzip+s3://`?

kylebarron:
Is that standard anywhere? Conversely, does S3 tell you the content type on request? Like in an http header

vincentsarago:
> Is that standard anywhere
it's standard in rasterio 🤔

kylebarron:
1. I'm not convinced this actually works with rasterio with gzip or zip files on s3. Have you tried it? The [docs](https://rasterio.readthedocs.io/en/stable/topics/datasets.html) only say it works with a local zip file.
2. Also, the rasterio docs link to the [Apache VFS spec here](https://commons.apache.org/proper/commons-vfs/filesystems.html#Zip_Jar_and_Tar).  But nowhere on that page does it mention `gzip+s3://` or anything similar as part of the spec. `+` doesn't even show up on the page!

	I think that spec defines 
	```
	gz://s3://bucket/key
	```
	but that's confusing.

Honestly the whole spec looks confusing.

I think the simplest method is to require that gzipped files end in `.gz`. That's pretty universally accepted, and combining that with a `gzip` arg should be enough for edge cases where a user wants to force compression/decompression.

vincentsarago:
> I think the simplest method is to require that gzipped files end in .gz. That's pretty universally accepted, and combining that with a gzip arg should be enough for edge cases where a user wants to force compression/decompression.

well in my case I'd love to save/read mosaic on S3 but I don't want to save them with extensions so I found that doing `gipz+s3://my-bucket/mosaic` could be convenient, because right now we don't have a way to force the gzip compression/decompression without the `.gz` at the end. We can revisit that later but wanted to throw the idea

kylebarron:
> but I don't want to save them with extensions

Is there a reason why saving them with extensions is problematic?

vincentsarago:
> Is there a reason why saving them with extensions is problematic?

not specifics

PR #56, opened 2020-04-22
Quadkey options

vincentsarago:
replace #41


cogeo_mosaic/utils.py
```diff
@@ -237,7 +240,7 @@ def create_mosaic(
 
         maxzoom = max(maxzoom)
 
-    quadkey_zoom = minzoom
+    quadkey_zoom = quadkey_zoom or minzoom
```
vincentsarago, line 243: @kylebarron I think this is the only change between your PR and mine

cogeo_mosaic/utils.py
```diff
@@ -177,6 +177,7 @@ def create_mosaic(
     dataset_list: Sequence[str],
     minzoom: int = None,
     maxzoom: int = None,
+    quadkey_zoom: int = None,
```
vincentsarago, line 180: Note there is a potential risque when a user will pass `quadkey_zoom` and `version==0.0.1`. But that's only on the API side (not on the CLI) so it's unlikely.

cogeo_mosaic/utils.py
kylebarron: I think
```py
quadkey_zoom = quadkey_zoom or minzoom
```
is simpler than
```py
    if quadkey_zoom is None:
        quadkey_zoom = minzoom
```

vincentsarago: I know that's why I write it!

cogeo_mosaic/utils.py
vincentsarago: will be taken care of in #57

setup.py
```diff
@@ -12,7 +12,7 @@
     "requests",
     "rio-cogeo>=1.1.0",
     "rio-tiler>=2.0a4",
-    "rio-tiler-mosaic>=0.0.1dev3",
+    "rio-tiler-mosaic>=0.0.1dev4",
```
vincentsarago, line 15: FYI, `0.0.1dev3` forced rio-tiler~1.2

PR #59, opened 2020-04-24
Support custom asset filter in inner create_mosaic function

kylebarron:
This PR allows a user to pass a custom `asset_filter` function, which receives the tile and all overlapping assets. This allows a user to rewrite the logic for which assets to keep for a given tile, without needing to duplicate any provided code in the package.

This is on top of #46 , so it looks messier than it is. The gist is just here:
https://github.com/developmentseed/cogeo-mosaic/compare/master...kylebarron:custom-optimizer?expand=1#diff-ee157dd7e791c97cb4e3e8ed1dab608bR176-R188


cogeo_mosaic/create.py
```diff
@@ -123,22 +173,11 @@ def create_mosaic(
...
-        if minimum_tile_cover is not None or tile_cover_sort:
-            dataset = _filter_and_sort(
-                tile_geom,
-                dataset,
-                minimum_cover=minimum_tile_cover,
-                sort_cover=tile_cover_sort,
-            )
+        intersections = [features[idx] for idx in intersections_idx]
+
+        dataset = asset_filter(tile, intersections, intersections_geoms, **kwargs)
```
kylebarron, line 178: @vincentsarago aka here, this is a function that takes in 

- `tile` (`mercantile.Tile`)
- `dataset` (List[Dict])
- `geoms` (pygeos polygons)
- `kwargs` any other parameters passed to the filter function.

This means that a user can easily override this function

PR #60, opened 2020-04-24
WIP Create mosaic class version

kylebarron:


cogeo_mosaic/model.py
```diff
@@ -24,3 +31,169 @@ class MosaicJSON(BaseModel):
...
+                    sort_cover=tile_cover_sort,
+                )
+
+            if dataset and maximum_items_per_tile:
+                dataset = dataset[0:maximum_items_per_tile]
+
+            if dataset:
+                mosaic_definition["tiles"][quadkey] = [f["identifier"] for f in dataset]
+
+        return MosaicJSON.from_dict(mosaic_definition)
```
vincentsarago, line 199: I think going directly with `return cls(**mosaic_definition)` is fine here

cogeo_mosaic/model.py
```diff
@@ -24,3 +31,169 @@ class MosaicJSON(BaseModel):
     bounds: List[float] = Field([-180, -90, 180, 90])
     center: Optional[Tuple[float, float, int]]
     tiles: Dict[str, List[str]]
+
+    @classmethod
+    def from_dict(cls, mosaic_def: Dict):
+        return MosaicJSON(**dict(mosaic_def))
```
vincentsarago, line 37: if it's already a dict you don't need `dict(mosaic_def)`?

cogeo_mosaic/model.py
```diff
@@ -24,3 +31,169 @@ class MosaicJSON(BaseModel):
...
+        datatype = {feat["properties"]["datatype"] for feat in features}
+        if len(datatype) > 1:
+            raise Exception("Dataset should have the same data type")
+
+        return MosaicJSON.from_features(
+            features, minzoom=minzoom, maxzoom=maxzoom, quiet=quiet, **kwargs
+        )
+
+    @classmethod
+    def _create_mosaic(
```
vincentsarago, line 101: I think this could be `from_feature` directly.

cogeo_mosaic/model.py
kylebarron: I was looking at the Geopandas source, and I got the `Classname(object)` syntax from them: https://github.com/geopandas/geopandas/blob/f13dafb881eaaf2bbf296e6f12189cc2c21a4216/geopandas/geodataframe.py#L387

cogeo_mosaic/model.py
kylebarron: I know, but it doesn't hurt to coerce

cogeo_mosaic/model.py
kylebarron: I feel that it's better to split it up for composability. It's a nice logical separation

cogeo_mosaic/model.py
vincentsarago: well it just doesn't make sense to me to re-dict a dict ;-)

cogeo_mosaic/model.py
geospatial-jeff: I prefer the `cls` constructor.  Regardless you don't need `from_dict`.

kylebarron: Well it isn't _wrong_ at least, since it works

PR #61, opened 2020-04-24
Mosaic class

vincentsarago:
overtake #60 #46 #59  

## To Do
- [x] review the implementation of the MosaicClass in respect of #60 #46 #59  
- [x] refactor update function
- [x] add `from stac` cli methods


cogeo_mosaic/mosaic.py
```diff
@@ -0,0 +1,241 @@
...
+        if minimum_tile_cover or tile_cover_sort:
+            tile_geom = polygons(mercantile.feature(tile)["geometry"]["coordinates"][0])
+            int_pcts = _intersect_percent(tile_geom, geoms)
+
+            if minimum_tile_cover:
+                indices = [ind for ind in indices if int_pcts[ind] > minimum_tile_cover]
+
+            if tile_cover_sort:
+                # https://stackoverflow.com/a/9764364
+                indices, _ = zip(*sorted(zip(indices, int_pcts), reverse=True))
```
vincentsarago, line 62: @kylebarron I added `reverse=True` here to match existent function

cogeo_mosaic/mosaic.py
```diff
@@ -0,0 +1,241 @@
...
+        minimum_tile_cover=None,
+        tile_cover_sort=False,
+        maximum_items_per_tile: Optional[int] = None,
+    ):
+        """Filter and/or sort dataset per intersection coverage."""
+        indices = list(range(len(dataset)))
+
+        if minimum_tile_cover or tile_cover_sort:
+            tile_geom = polygons(mercantile.feature(tile)["geometry"]["coordinates"][0])
+            int_pcts = _intersect_percent(tile_geom, geoms)
```
vincentsarago, line 55: we needed to pass list of `geoms` to `_intersect_percent` function

cogeo_mosaic/mosaic.py
```diff
@@ -0,0 +1,241 @@
...
+            tile_geom = polygons(mercantile.feature(tile)["geometry"]["coordinates"][0])
+
+            # Find intersections from rtree
+            intersections_idx = sorted(tree.query(tile_geom, predicate="intersects"))
+            if len(intersections_idx) == 0:
+                continue
+
+            intersect_dataset, intersect_geoms = zip(
+                *[(features[idx], dataset_geoms[idx]) for idx in intersections_idx]
+            )
```
vincentsarago, line 151: one line is better than 2 lines

cogeo_mosaic/mosaic.py
kylebarron: Oh I forgot about that. Reverse sorts highest to lowest then?

cogeo_mosaic/mosaic.py
kylebarron: I think that's what I was trying to do but failed

cogeo_mosaic/mosaic.py
kylebarron: Not if it's less readable 😉

cogeo_mosaic/mosaic.py
vincentsarago: we reverse we sort % intersection from highest to lowest

cogeo_mosaic/mosaic.py
kylebarron: I do think this is more readable as two lines. And the length of `intersections_idx` is very small

cogeo_mosaic/mosaic.py
```diff
@@ -0,0 +1,241 @@
...
+        cls,
+        urls: Sequence[str],
+        minzoom: Optional[int] = None,
+        maxzoom: Optional[int] = None,
+        max_threads: int = 20,
+        quiet: bool = True,
+        **kwargs,
+    ):
+        """
+        Create mosaicjson from url of COGs.
```
kylebarron, line 171: `urls`. Aka we're passing many urls that each point to one COG, not the reverse.

cogeo_mosaic/mosaic.py
```diff
@@ -0,0 +1,241 @@
...
+    def from_features(cls, features: Sequence[Dict], **kwargs):
+        """
+        Create mosaicjson from a set of GeoJSON Features.
+
+        Attributes
+        ----------
+        features: list, required
+            List of GeoJSON features.
+        kwargs: any
+            Options forwarded to MosaicJSON.from_features
```
kylebarron, line 233: Forwarded to `_create_mosaic`

cogeo_mosaic/mosaic.py
```diff
@@ -0,0 +1,241 @@
...
+        datatype = {feat["properties"]["datatype"] for feat in features}
+        if len(datatype) > 1:
+            raise Exception("Dataset should have the same data type")
+
+        return cls._create_mosaic(
+            features, minzoom=minzoom, maxzoom=maxzoom, quiet=quiet, **kwargs
+        )
+
+    @classmethod
+    def from_features(cls, features: Sequence[Dict], **kwargs):
```
kylebarron, line 224: You really can just set everything here as ``**kwargs``. Just include the full list of attributes in the docstring.

kylebarron: Nevermind, `features` allows it to be positional.

vincentsarago:
Alright I'm going to update the docs and then we'll be able to merge this

vincentsarago:
Note: 
Latest commit refactored a bit how we handle `Update`. Basically the update method for `File` and `S3` backend should be able to `write` at the end, while it shouldn't for dynamoDB backend. This adds a bit of code duplication but I think it's for the best

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -52,9 +54,76 @@ def write(self):
...
+        ]
+
+        nxmin, nymin, nxmax, nymax = new_mosaic.bounds
+        oxmin, oymin, oxmax, oymax = self.mosaic_def.bounds
+        bounds = [
+            min(nxmin, oxmin),
+            min(nymin, oymin),
+            max(nxmax, oxmax),
+            max(nymax, oymax),
+        ]
```
vincentsarago, line 123: 🤦

PR #63, opened 2020-04-24
Composable CLI

kylebarron:
This is a small refactor to have a really clean, composable CLI:
```
> echo "tests/fixtures/cog1.tif" | cogeo-mosaic footprint | cogeo-mosaic create
Get quadkey list for zoom: 7
Feed Quadkey index
{"mosaicjson": "0.0.2", "version": "1.0.0", "minzoom": 7, "maxzoom": 9, "quadkey_zoom": 7, "bounds": [-75.98703377413767, 44.94205659982972, -72.90811027838137, 47.09314422696482], "center": [-74.44757202625952, 46.01760041339727, 7], "tiles": {"0302300": ["tests/fixtures/cog1.tif"], "0302301": ["tests/fixtures/cog1.tif"], "0302310": ["tests/fixtures/cog1.tif"], "0302302": ["tests/fixtures/cog1.tif"], "0302303": ["tests/fixtures/cog1.tif"], "0302312": ["tests/fixtures/cog1.tif"], "0302320": ["tests/fixtures/cog1.tif"], "0302321": ["tests/fixtures/cog1.tif"], "0302330": ["tests/fixtures/cog1.tif"]}}
```

PR #64, opened 2020-04-24
add CLI to create a mosaic from GeoJSON features

vincentsarago:
This PR adds a CLI to create a mosaicJSON from a list of features of featureCollection

### DIFF
--- a/cogeo_mosaic/scripts/cli.py
+++ b/cogeo_mosaic/scripts/cli.py
@@ -7,6 +7,7 @@ from pkg_resources import iter_entry_points
 
 from click_plugins import with_plugins
 import click
+import cligj
 
 from cogeo_mosaic import version as cogeo_mosaic_version
 from cogeo_mosaic.utils import get_footprints, update_mosaic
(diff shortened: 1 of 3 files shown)


cogeo_mosaic/scripts/cli.py
```diff
@@ -92,6 +93,68 @@ def create(
...
+    quadkey_zoom,
+    min_tile_cover,
+    tile_cover_sort,
+    quiet,
+):
+    """Create mosaic definition file."""
+    mosaicjson = MosaicJSON.from_features(
+        list(features),
+        minzoom,
+        maxzoom,
```
kylebarron, line 143: Just pass these as keyword args so that you don't need to add more positional arguments to `from_features`

vincentsarago: nope because thoses are required in the actual model.

kylebarron: Just use this: https://github.com/developmentseed/cogeo-mosaic/pull/63/files#diff-7c1d4d4b7c5cc98334549838d79b4534R109-R136

Don't make them required from the command line, and tell people that if the properties don't exist, an error will be raised.

vincentsarago: I don't think it's a good way for now. Happy to review this later

PR #65, opened 2020-04-27
Update mosaic

vincentsarago:
This PR aims to add the `update` method for all backends. 

To be able to handler update for different backends and by maximizing code re-use I had to add 2 privates methods in the base class: 
- `_update_quadkey`: update quadkey dataset
- `_update_metadata`: update bounds, center and version metadata

TBD: the `update` method introduces a new `concept` with the `on_top` option. This is basically to let the user happens the new data on top or on the bottom of the dataset list. 

The problem with this `on_top` and with looping through the datafeatures, is that when a user will pass a list (e.g `1.tiff, 2.tiff`) with `on_top` we will end up with `[2.tif, 1.tif, alreadyinlist.tif]` so we might want to reverse the order of the input list when `on_top=True`. This feature will also need more documentation 

### DIFF
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -1,6 +1,6 @@
 """cogeo_mosaic.backend.base: base Backend class."""
 
-from typing import Any, Callable, Dict, List, Sequence
+from typing import Callable, Dict, List, Sequence
 
 import abc
 from contextlib import AbstractContextManager
@@ -17,6 +17,7 @@ from cogeo_mosaic.backends.utils import get_hash
 class BaseBackend(AbstractContextManager):
     """Base Class for cogeo-mosaic backend storage."""
 
+    path: str
     mosaic_def: MosaicJSON
 
     @abc.abstractmethod
@@ -66,16 +67,6 @@ class BaseBackend(AbstractContextManager):
     def write(self):
         """Upload new MosaicJSON to backend."""
 
-    @abc.abstractmethod
-    def update(
-        self,
-        features: Sequence[Dict],
-        accessor: Callable = DEFAULT_ACCESSOR,
-        overwrite: bool = False,
-        **kwargs: Any,
-    ):
-        """Update existing MosaicJSON on backend."""
-
     def _update_quadkey(self, quadkey: str, dataset: List[str]):
         """Update quadkey list."""
         self.mosaic_def.tiles[quadkey] = dataset
@@ -90,10 +81,10 @@ class BaseBackend(AbstractContextManager):
             self.mosaic_def.minzoom,
         )
 
-    def _update(
+    def update(
         self,
         features: Sequence[Dict],
-        accessor: Callable,
+        accessor: Callable = DEFAULT_ACCESSOR,
         add_first: bool = True,
         **kwargs,
     ):
@@ -146,4 +137,6 @@ class BaseBackend(AbstractContextManager):
         )
         self._update_metadata(bounds, version)
 
+        if self.path:
+            self.write()
         return
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -77,19 +77,13 @@ class BaseBackend(AbstractContextManager):
             self.mosaic_def.minzoom,
         )
 
-    def update(
-        self,
-        features: Sequence[Dict],
-        add_first: bool = True,
-        mosaic_class=MosaicJSON,
-        **kwargs,
-    ):
+    def update(self, features: Sequence[Dict], add_first: bool = True, **kwargs):
         """Update existing MosaicJSON on backend."""
         version = list(map(int, self.mosaic_def.version.split(".")))
         version[-1] += 1
-        version = ".".join(map(str, version))
+        new_version = ".".join(map(str, version))
 
-        new_mosaic = mosaic_class.from_features(
+        new_mosaic = self.mosaic_def.from_features(
             features,
             self.mosaic_def.minzoom,
             self.mosaic_def.maxzoom,
@@ -112,7 +106,7 @@ class BaseBackend(AbstractContextManager):
             max(nxmax, oxmax),
             max(nymax, oymax),
         ]
-        self._update_metadata(bounds, version)
+        self._update_metadata(bounds, new_version)
 
         # We only write if path is set
         if self.path:
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -108,38 +108,35 @@ class BaseBackend(AbstractContextManager):
             [feat["geometry"]["coordinates"][0] for feat in features]
         )
 
-        for idx, feature in enumerate(features):
-            tiles = burntiles.burn([feature], self.quadkey_zoom)
-            tiles = [mercantile.Tile(*tile) for tile in tiles]
+        tiles = burntiles.burn(features, self.quadkey_zoom)
+        tiles = [mercantile.Tile(*tile) for tile in tiles]
 
-            tree = STRtree([dataset_geoms[idx]])
+        tree = STRtree(dataset_geoms)
 
-            for tile in tiles:
-                quadkey = mercantile.quadkey(tile)
-                tile_geom = polygons(
-                    mercantile.feature(tile)["geometry"]["coordinates"][0]
-                )
+        for tile in tiles:
+            quadkey = mercantile.quadkey(tile)
+            tile_geom = polygons(mercantile.feature(tile)["geometry"]["coordinates"][0])
 
-                intersections_idx = sorted(
-                    tree.query(tile_geom, predicate="intersects")
-                )
-                if len(intersections_idx) == 0:
-                    continue
+            # Find intersections from rtree
+            intersections_idx = sorted(tree.query(tile_geom, predicate="intersects"))
+            if len(intersections_idx) == 0:
+                continue
 
-                intersect_dataset, intersect_geoms = zip(
-                    *[(features[idx], dataset_geoms[idx]) for idx in intersections_idx]
-                )
+            intersect_dataset, intersect_geoms = zip(
+                *[(features[idx], dataset_geoms[idx]) for idx in intersections_idx]
+            )
 
-                dataset = self.mosaic_def._filter(
-                    tile, intersect_dataset, intersect_geoms, **kwargs
-                )
-                new_assets = [accessor(f) for f in dataset]
+            dataset = self.mosaic_def._filter(
+                tile, intersect_dataset, intersect_geoms, **kwargs
+            )
 
-                assets = self.tile(*tile)
-                assets = [*new_assets, *assets] if add_first else [*assets, *new_assets]
-                self._update_quadkey(quadkey, assets)
+            new_assets = [accessor(f) for f in dataset]
 
-                updated_quadkeys.add(tile)
+            assets = self.tile(*tile)
+            assets = [*new_assets, *assets] if add_first else [*assets, *new_assets]
+            self._update_quadkey(quadkey, assets)
+
+            updated_quadkeys.add(tile)
 
         bounds = self.mosaic_def.bounds
         minimumTile = mercantile.tile(bounds[0], bounds[3], self.quadkey_zoom)
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -1,16 +1,12 @@
 """cogeo_mosaic.backend.base: base Backend class."""
 
-from typing import Callable, Dict, List, Sequence
+from typing import Dict, List, Sequence
 
 import abc
 from contextlib import AbstractContextManager
 
 import mercantile
-from pygeos import STRtree, polygons
-from supermercado import burntiles
-
-from cogeo_mosaic.utils import tiles_to_bounds
-from cogeo_mosaic.mosaic import MosaicJSON, DEFAULT_ACCESSOR
+from cogeo_mosaic.mosaic import MosaicJSON
 from cogeo_mosaic.backends.utils import get_hash
 
 
@@ -84,8 +80,8 @@ class BaseBackend(AbstractContextManager):
     def update(
         self,
         features: Sequence[Dict],
-        accessor: Callable = DEFAULT_ACCESSOR,
         add_first: bool = True,
+        mosaic_class=MosaicJSON,
         **kwargs,
     ):
         """Update existing MosaicJSON on backend."""
@@ -93,50 +89,33 @@ class BaseBackend(AbstractContextManager):
         version[-1] += 1
         version = ".".join(map(str, version))
 
-        updated_quadkeys = set()
-
-        dataset_geoms = polygons(
-            [feat["geometry"]["coordinates"][0] for feat in features]
+        new_mosaic = mosaic_class.from_features(
+            features,
+            self.mosaic_def.minzoom,
+            self.mosaic_def.maxzoom,
+            quadkey_zoom=self.quadkey_zoom,
+            **kwargs,
         )
-
-        tiles = burntiles.burn(features, self.quadkey_zoom)
-        tiles = [mercantile.Tile(*tile) for tile in tiles]
-
-        tree = STRtree(dataset_geoms)
-
-        for tile in tiles:
-            quadkey = mercantile.quadkey(tile)
-            tile_geom = polygons(mercantile.feature(tile)["geometry"]["coordinates"][0])
-
-            # Find intersections from rtree
-            intersections_idx = sorted(tree.query(tile_geom, predicate="intersects"))
-            if len(intersections_idx) == 0:
-                continue
-
-            intersect_dataset, intersect_geoms = zip(
-                *[(features[idx], dataset_geoms[idx]) for idx in intersections_idx]
-            )
-
-            dataset = self.mosaic_def._filter(
-                tile, intersect_dataset, intersect_geoms, **kwargs
-            )
-
-            new_assets = [accessor(f) for f in dataset]
-
+        for quadkey, new_assets in new_mosaic.tiles.items():
+            tile = mercantile.quadkey_to_tile(quadkey)
             assets = self.tile(*tile)
             assets = [*new_assets, *assets] if add_first else [*assets, *new_assets]
+
+            # add custom sorting algorithm (e.g based on path name)
             self._update_quadkey(quadkey, assets)
 
-            updated_quadkeys.add(tile)
-
-        bounds = self.mosaic_def.bounds
-        minimumTile = mercantile.tile(bounds[0], bounds[3], self.quadkey_zoom)
-        maximumTile = mercantile.tile(bounds[2], bounds[1], self.quadkey_zoom)
-        bounds = tiles_to_bounds(
-            [t for t in updated_quadkeys] + [minimumTile, maximumTile]
-        )
+        nxmin, nymin, nxmax, nymax = new_mosaic.bounds
+        oxmin, oymin, oxmax, oymax = self.mosaic_def.bounds
+        bounds = [
+            min(nxmin, oxmin),
+            min(nymin, oymin),
+            max(nxmax, oxmax),
+            max(nymax, oymax),
+        ]
         self._update_metadata(bounds, version)
 
+        # We only write if path is set
         if self.path:
             self.write()
+
         return
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -1,10 +1,15 @@
 """cogeo_mosaic.backend.base: base Backend class."""
 
-from typing import Dict, List
+from typing import Callable, Dict, List, Sequence
 
 import abc
 from contextlib import AbstractContextManager
 
+import mercantile
+from pygeos import STRtree, polygons
+from supermercado import burntiles
+
+from cogeo_mosaic.utils import tiles_to_bounds
 from cogeo_mosaic.mosaic import MosaicJSON
 from cogeo_mosaic.backends.utils import get_hash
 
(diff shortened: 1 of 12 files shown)


cogeo_mosaic/backends/base.py
```diff
@@ -64,3 +69,83 @@ def write(self):
...
+
+    def _update_metadata(self, bounds: List[float], version: str):
+        """Update bounds and center."""
+        self.mosaic_def.version = version
+        self.mosaic_def.bounds = bounds
+        self.mosaic_def.center = (
+            (bounds[0] + bounds[2]) / 2,
+            (bounds[1] + bounds[3]) / 2,
+            self.mosaic_def.minzoom,
+        )
```
vincentsarago, line 85: @kylebarron is a Tuple ok for dynamoDB ?

kylebarron: I think so, but you need to coerce the numeric values to `Decimal`.

cogeo_mosaic/backends/base.py
```diff
@@ -64,3 +69,83 @@ def write(self):
...
+                if len(intersections_idx) == 0:
+                    continue
+
+                intersect_dataset, intersect_geoms = zip(
+                    *[(features[idx], dataset_geoms[idx]) for idx in intersections_idx]
+                )
+
+                dataset = self.mosaic_def._filter(
+                    tile, intersect_dataset, intersect_geoms, **kwargs
+                )
```
vincentsarago, line 134: for now on `minimum_tile_cover` is a valid kwargs I'll refactor a bit to first read the tile dataset ?

cogeo_mosaic/backends/file.py
```diff
@@ -53,9 +53,17 @@ def write(self, gzip: bool = None):
...
+        features: Sequence[Dict],
+        accessor: Callable,
+        overwrite: bool = False,
+        **kwargs: Any
+    ):
         """Update the mosaicjson document."""
-        raise NotImplementedError
+        self._update(features, accessor, **kwargs)
+        if overwrite:
+            self.write()
```
vincentsarago, line 66: `backend.update` handles mosaicjson update and overwriting.

cogeo_mosaic/backends/http.py
```diff
@@ -45,9 +46,17 @@ def write(self):
...
+        features: Sequence[Dict],
+        accessor: Callable,
+        overwrite: bool = False,
+        **kwargs: Any
+    ):
         """Update the mosaicjson document."""
-        raise NotImplementedError
+        self._update(features, accessor, **kwargs)
+        if overwrite:
+            warnings.warn("Overwrite is not possible for http backend")
```
vincentsarago, line 59: so if overwrite in set to `True` we will update the mosaic internally but with the actual CLI, it won't be printed

cogeo_mosaic/scripts/cli.py
```diff
@@ -150,31 +150,40 @@ def create_from_features(
...
     help="threads",
 )
-def update(input_files, input_mosaic, output, min_tile_cover, threads):
+def update(input_files, input_mosaic, min_tile_cover, on_top, overwrite, threads):
     """Update mosaic definition file."""
+    if input_mosaic.startswith("dynamodb://") and not overwrite:
+        raise Exception(
+            "DynamoDB hosted mosaic has to be overwriten on update. Please use `--overwrite`"
+        )
+
```
vincentsarago, line 173: we could also add other check to make overwrite set to `False` for http backend

cogeo_mosaic/scripts/cli.py
```diff
@@ -150,31 +150,40 @@ def create_from_features(
 @cogeo_cli.command(short_help="Create mosaic definition from list of files")
 @click.argument("input_files", type=click.File(mode="r"), default="-")
 @click.argument("input_mosaic", type=click.Path())
-@click.option("--output", "-o", type=click.Path(exists=False), help="Output file name")
 @click.option("--min-tile-cover", type=float, help="Minimum % overlap")
+@click.option(
+    "--on-top/--on-bottom",
```
vincentsarago, line 155: need better words and phrasing

cogeo_mosaic/mosaic.py
```diff
@@ -138,7 +138,7 @@ def _create_mosaic(
         tree = STRtree(dataset_geoms)
 
         for tile in tiles:
-            quadkey = mercantile.quadkey(tile)
+            quadkey = str(mercantile.quadkey(tile))
```
vincentsarago, line 141: FIX, this should be a string (per the specification)

cogeo_mosaic/backends/base.py
```diff
@@ -64,3 +69,83 @@ def write(self):
...
+        else:
+            version = "1.0.0"
+
+        updated_quadkeys = set()
+
+        dataset_geoms = polygons(
+            [feat["geometry"]["coordinates"][0] for feat in features]
+        )
+
+        for idx, feature in enumerate(features):
```
vincentsarago, line 109: do we want a progressbar ?

cogeo_mosaic/backends/base.py
vincentsarago: oh I remember why I don't read the `assets` before, it's because we don't have the geometry of them, so we can not apply any other filtering (e.g `minimum_tile_cover`, `tile_cover_sort`) on the full list (old + new assets)

cogeo_mosaic/mosaic.py
kylebarron: It already is a string.

cogeo_mosaic/backends/base.py
```diff
@@ -64,3 +69,82 @@ def write(self):
     @abc.abstractmethod
     def update(self):
```
geospatial-jeff, line 70: Update signature to match children.

```python
    def update(
        self,
        features: Sequence[Dict],
        accessor: Callable,
        overwrite: bool = False,
        **kwargs: Any
    ):
```

cogeo_mosaic/backends/base.py
```diff
@@ -64,3 +69,82 @@ def write(self):
...
         """Update existing MosaicJSON on backend."""
+
+    def _update_quadkey(self, quadkey: str, dataset: List[str]):
```
geospatial-jeff, line 73: Is there a situation where the user wants to update a single quadkey?  If not I don't think we need this method.

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -52,9 +52,35 @@ def write(self):
         items = self._create_items()
         self._write_items(items)
 
-    def update(self):
+    def _update_quadkey(self, quadkey: str, dataset: List[str]):
+        """Update quadkey list."""
+        self.table.put_item(Item={"quadkey": quadkey, "assets": dataset})
+
+    def _update_metadata(self, bounds: List[float], version: str):
```
geospatial-jeff, line 59: Should this also update `self.mosaic_def`? to stay in sync with what is in dynamodb

cogeo_mosaic/backends/base.py
geospatial-jeff: Maybe also have `accessor` default to `DEFAULT_ACESSOR`

cogeo_mosaic/backends/base.py
```diff
@@ -64,3 +69,82 @@ def write(self):
...
+        """Update quadkey list."""
+        self.mosaic_def.tiles[quadkey] = dataset
+
+    def _update_metadata(self, bounds: List[float], version: str):
```
geospatial-jeff, line 77: Since we are updating the model directly it would be nice to re-validate on assignment using the `validate_assignment` [model config](https://pydantic-docs.helpmanual.io/usage/model_config/)

```python
class MosaicJSON(BaseModel):
    ...

    class Config:
        validate_asignment = True
```

cogeo_mosaic/backends/base.py
vincentsarago: > Maybe also have accessor default to DEFAULT_ACESSOR

I was hesitant but yes

cogeo_mosaic/backends/base.py
vincentsarago: this was made to accommodate DyanmoDB where we need to update each quadkey

cogeo_mosaic/backends/base.py
geospatial-jeff: Progressbar would be nice for CLI, as long as you can turn it off.  Personally I like the rio-cogeo usage https://github.com/cogeotiff/rio-cogeo/blob/master/rio_cogeo/cogeo.py#L70

cogeo_mosaic/backends/base.py
kylebarron: Definitely something you need to test.

kylebarron:
The big issue I have with the current setup is that you have potentially many unnecessary read/writes for DynamoDB. You have a double loop here by looping over tiles for each feature:

```py
        for idx, feature in enumerate(features):
            for tile in tiles:
                self._update_quadkey(quadkey, assets)
```

If those features are overlapping, 

1. you'll be reading and updated data you _just_ uploaded
2. it'll take longer because the slowest part of the update is the dynamodb read/writes over the network

So instead, I'd suggest having a first loop to find overlapping tiles and associate features with those tiles. Something like:
```py
tiles_dict = {}
for idx, feature in enumerate(features):
	tiles = find_tiles(feature)
	for tile in tiles:
		# Initialize
		tiles_dict[tile] = tiles_dict.get(tile, set())
	
		tiles_dict[tile].add(idx)
```
Then `tiles_dict` is a mapping from mercator tiles to features, so that you can do one update for each tile.

Then you can have a second loop:
```py
for tile, features in tiles_dict.items():
	# get tile from mosaic
	# Update tile with all features
	# Put tile back
```

cogeo_mosaic/backends/base.py
kylebarron: Yeah I think `_filter` here is weird. Because in general `_filter` expects to take the entire dataset. I'm not sure whether it really applies here.

Also, since this is only applied on _new_ assets, if you pass `maximum_items_per_tile=5`, that restricts to 5 _new_ assets, so your mosaic could grow by 5 items per tile every time you call `update`.

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -52,9 +52,35 @@ def write(self):
...
+        self,
+        features: Sequence[Dict],
+        accessor: Callable,
+        overwrite: bool = False,
+        **kwargs: Any
+    ):
         """Update the mosaicjson document."""
-        raise NotImplementedError
+        if not overwrite:
+            raise Exception("Overwrite has to be set to True")
```
kylebarron, line 81: I don't understand this. If anything, it _can't_ be `True`, because to me an "overwrite" is equivalent to "delete a file and write a new file". The only way for that to happen with DynamoDB is to delete the table and create it again, or delete every individual quadkey.

So essentially only an "update" or "append" is possible with dynamodb.

cogeo_mosaic/backends/file.py
kylebarron: So if `overwrite` is `False`, that means update the internal mosaicjson document but don't write it to the backend?

cogeo_mosaic/scripts/cli.py
kylebarron: Like `--add-first`, `--add-last`

cogeo_mosaic/scripts/cli.py
```diff
@@ -150,31 +150,40 @@ def create_from_features(
...
+)
+@click.option("--overwrite", help="overwrite mosaic.", is_flag=True)
 @click.option(
     "--threads",
     type=int,
     default=lambda: os.environ.get("MAX_THREADS", multiprocessing.cpu_count() * 5),
     help="threads",
 )
-def update(input_files, input_mosaic, output, min_tile_cover, threads):
+def update(input_files, input_mosaic, min_tile_cover, on_top, overwrite, threads):
```
kylebarron, line 167: So `min_tile_cover` only applies to new assets?

cogeo_mosaic/backends/file.py
vincentsarago: yes

cogeo_mosaic/scripts/cli.py
vincentsarago: yes

cogeo_mosaic/backends/dynamodb.py
vincentsarago: > So essentially only an "update" or "append" is possible with dynamodb.

I think I had to add the overwrite option to be consistent with the other backend https://github.com/developmentseed/cogeo-mosaic/blob/updateMosaic/cogeo_mosaic/scripts/cli.py#L182

cogeo_mosaic/backends/base.py
vincentsarago: yeah basically this is why I only expose `minimum_tile_cover` because we cannot have the whole stack of dataset. This is totally weird but has to do with the new implementation

cogeo_mosaic/backends/dynamodb.py
geospatial-jeff: As written it does overwrite https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.put_item

geospatial-jeff:
> The big issue I have with the current setup is that you have potentially many unnecessary read/writes for DynamoDB.

Maybe do batch writes in chunks as described here https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.batch_write_item

cogeo_mosaic/backends/dynamodb.py
kylebarron: Right but that overwrites the _quadkey_, not the _mosaic_. With other backends, overwriting always overwrites the entire _file_, so this behavior isn't consistent.

kylebarron:
> Maybe do batch writes in chunks as described here

Doing batched writes is good, but doesn't solve the issue, because you could still be reading/writing the same quadkey multiple times, unnecessarily. But in the second loop I described above, when you have the entire list of quadkeys you plan to update, doing a batch read and then a batch write might be better than individual read/writes. I'm not sure

cogeo_mosaic/backends/dynamodb.py
vincentsarago: > I don't understand this. If anything, it can't be True, because to me an "overwrite" is equivalent to "delete a file and write a new file".

yeah to me overwrite, mean editing a file in-place.

> With other backends, overwriting always overwrites the entire file, so this behavior isn't consistent.

Yeah if you have better phrasing that works for all the backend I'm totally open to change it!

vincentsarago: >  If anything, it can't be True

well I'm not sure to understand, for DynamoDB backend `overwrite` has to be set to True.

geospatial-jeff: Other backends use a file as a representation of the mosaic.  When the update method is called with `overwrite=True` it replace the file with a new one containing the updated metadata and quadkeys.  Dynamo is different representation but same behavior ([overwrite metadata](https://github.com/developmentseed/cogeo-mosaic/blob/updateMosaic/cogeo_mosaic/backends/dynamodb.py#L70) and [overwrite quadkeys](https://github.com/developmentseed/cogeo-mosaic/blob/updateMosaic/cogeo_mosaic/backends/dynamodb.py#L57)).

kylebarron: > yeah to me overwrite, mean editing a file in-place.

I think we have differing definitions. With a normal file, there's no such think as an "update in place", because you read the entire file and overwrite _the entire file_. It's not so much an update as it is deleting everything and writing new content.

Since DynamoDB is a database, my point is that it doesn't modify quadkeys that it doesn't read, so I think `overwrite` is a bit of a misnomer.

kylebarron: Even using what I think is your definition of `overwrite`, when `overwrite` is `False`, it updates the mosaic locally but not remotely. I don't see a reason why that isn't possible for DynamoDB?

vincentsarago: and my question was do you think of a word that will fit for all the backend ?

kylebarron: I need to think about it a bit more. At first glance, I would say that the `update` method _on the backend class_ should always commit the updates.

vincentsarago: yeah I think this comes down to, should we or should we not write the update! we will endup not having an update method for http backends but I'm fine with this

vincentsarago: > yeah I think this comes down to, `should we or should we not write the update!`If we choose to alway `write` we will end up not having an update method for http backends but I'm fine with this

@geospatial-jeff @kylebarron let me know what you think!

cogeo_mosaic/scripts/cli.py
```diff
@@ -152,8 +152,8 @@ def create_from_features(
 @click.argument("input_mosaic", type=click.Path())
 @click.option("--min-tile-cover", type=float, help="Minimum % overlap")
 @click.option(
-    "--on-top/--on-bottom",
-    help="happens dataset on top of the existing scenes.",
+    "--add-first/--add-last",
+    help="Happens dataset on top of the existing scenes.",
```
kylebarron, line 156: Appends

cogeo_mosaic/backends/base.py
```diff
@@ -62,5 +67,83 @@ def write(self):
...
+                tile, intersect_dataset, intersect_geoms, **kwargs
+            )
+
+            new_assets = [accessor(f) for f in dataset]
+
+            assets = self.tile(*tile)
+            assets = [*new_assets, *assets] if add_first else [*assets, *new_assets]
+            self._update_quadkey(quadkey, assets)
+
+            updated_quadkeys.add(tile)
```
vincentsarago, line 139: @kylebarron In fact I just had to copy code from `MosaicJSON._create_mosaic`. This should reduce the number of read/write

kylebarron: I was originally going to say that you could probably just let `_create_mosaic` take an optional argument, something like `existing_mosaic`, that defaults to `None`, but if it's a `dict`, you fill the new tiles into that.

That way you aren't duplicating this code, but I'm not sure it's worth it to reduce this one duplication.

vincentsarago: Well I think this is doable. We can create a mosaic with the new dataset and then merge the `tiles` dict together.

I'll do this when #66 get merged

vincentsarago:
```python
    def update(
        self,
        features: Sequence[Dict],
        add_first: bool = True,
        **kwargs,
    ):
        version = list(map(int, self.mosaic_def.version.split(".")))
        version[-1] += 1
        version = ".".join(map(str, version))

        new_mosaic = MosaicJSON.from_features(
            features,
            self.mosaic_def.minzoom,
            self.mosaic_def.maxzoom,
            quadkey_zoom=self.quadkey_zoom,
            **kwargs,
        )
        for quadkey, new_assets in new_mosaic.tiles.items():
            tile = mercantile.quadkey_to_tile(quadkey)
            assets = self.tile(*tile)
            assets = [*new_assets, *assets] if add_first else [*assets, *new_assets]

            # add custom sorting algorithm (e.g based on path name)
            self._update_quadkey(quadkey, assets)

        nxmin, nymin, nxmax, nymax = new_mosaic.bounds
        oxmin, oymin, oxmax, oymax = self.mosaic_def.bounds

        bounds = [
            min(nxmin, oxmin), min(nymin, oymin), max(nxmax, oxmax), max(nymax, oymax)
        ]
        self._update_metadata(bounds, version)

        # We only write is path is set
        if self.path:
            self.write()

        return
```

cogeo_mosaic/backends/base.py
```diff
@@ -61,6 +63,59 @@ def quadkey_zoom(self) -> int:
...
+            (bounds[0] + bounds[2]) / 2,
+            (bounds[1] + bounds[3]) / 2,
+            self.mosaic_def.minzoom,
+        )
+
+    def update(
+        self,
+        features: Sequence[Dict],
+        add_first: bool = True,
+        mosaic_class=MosaicJSON,
```
vincentsarago, line 84: I'm adding `mosaic_class` as an option but not sure it is needed ? How does one use a custom mosaic class (e.g custom filter method) with backends? 

Backends have `MosaicJSON` hardcoded https://github.com/developmentseed/cogeo-mosaic/blob/MosaicClass/cogeo_mosaic/backends/base.py#L15, should this be passed as an options ? cc @geospatial-jeff @kylebarron

cogeo_mosaic/backends/base.py
```diff
@@ -77,19 +77,13 @@ def _update_metadata(self, bounds: List[float], version: str):
             self.mosaic_def.minzoom,
         )
 
-    def update(
-        self,
-        features: Sequence[Dict],
-        add_first: bool = True,
-        mosaic_class=MosaicJSON,
```
vincentsarago: In fact we need the mosaic_class to be defined if we want to use custom class

cogeo_mosaic/backends/base.py
geospatial-jeff: Thats a good point, I hadn't thought of that.  I think the options are (1) have the user monkeypatch the `_filter` class method or (2) pass a custom `MosaicJSON` as you mentioned.

vincentsarago: alight I think https://github.com/developmentseed/cogeo-mosaic/pull/67 would make it work. 

MosaicJSON class shouldn't really need to be customized, only the `asset_filter`and the `accessor` need to be.

With #67, you can now pass the custom `asset_filter` directly to the update method.

vincentsarago: That being said I think it's still not perfect because you don't have the ability to apply filter on the whole stack of data (new data and old data). To be able to do this you'll need a customBackend with a custom `Update` method.

vincentsarago: maybe we can turn https://github.com/developmentseed/cogeo-mosaic/pull/65/files#diff-79db4f358a4b5404256fcb377a494e21R96 into another `Callable` 🤔

cogeo_mosaic/backends/base.py
vincentsarago: FYI, because we do `meta = json.loads(json.dumps(self.metadata), parse_float=Decimal)`  before uploading the metadata to dynamDB, the `bounds` will be stored as a `[ ]` but when loading to the mosaicJSON it will be asigned to Tuple as expected

PR #66, opened 2020-04-28
remove http update method

vincentsarago:
This PR force `write` on `update` and thus disable update for HTTP Backends.

Changes:
- add `self.path` in mosaic to store url or path
- removes the unnecessary `_update`

### DIFF
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -1,6 +1,6 @@
 """cogeo_mosaic.backend.base: base Backend class."""
 
-from typing import Any, Callable, Dict, List, Sequence
+from typing import Callable, Dict, List, Sequence
 
 import abc
 from contextlib import AbstractContextManager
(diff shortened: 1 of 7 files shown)


cogeo_mosaic/backends/base.py
```diff
@@ -146,4 +137,6 @@ def _update(
         )
         self._update_metadata(bounds, version)
 
+        if self.path:
+            self.write()
```
vincentsarago, line 141: This is mostly to support `virtual` mosaic when doing 
```python
with MosaicBackend(None, mosaic_def=mymosaic) as mosaic:
     mosaic.update([1.tif])
```

kylebarron: Does every backend have a `path` attribute?

vincentsarago: yes starting with this PR

PR #67, opened 2020-04-30
refactor mosaic filter

vincentsarago:
This pr follow the original idea from Kyle in #59 but implemented with the new mosaic class. 

This PR should enable user to pass custom filter in any update/create method without the need to create custom Mosaic Class. 

```python
from cogeo_mosaic.mosaic import MosaicJSON, default_filter

def _filter_and_sort(*args, **kwargs):
    dataset = default_filter(*args, **kwargs)
    return sorted(dataset, key=lambda x: x["properties"]["path"], reverse=True)

mosaic = MosaicJSON.from_urls(assets, asset_filter=_filter_and_sort, quiet=False)
```

cc @geospatial-jeff @kylebarron

### DIFF
--- a/cogeo_mosaic/mosaic.py
+++ b/cogeo_mosaic/mosaic.py
@@ -14,7 +14,37 @@ from cogeo_mosaic.utils import get_footprints, _intersect_percent
 from pydantic import BaseModel, Field
 
 
-DEFAULT_ACCESSOR: Callable = lambda feature: feature["properties"]["path"]
+def default_accessor(feature: Dict):
+    """Return specific feature identifier."""
+    return feature["properties"]["path"]
+
+
+def default_filter(
+    tile: mercantile.Tile,
+    dataset: Sequence[Dict],
+    geoms: Sequence[polygons],
+    minimum_tile_cover=None,
+    tile_cover_sort=False,
+    maximum_items_per_tile: Optional[int] = None,
+) -> List:
+    """Filter and/or sort dataset per intersection coverage."""
+    indices = list(range(len(dataset)))
+
+    if minimum_tile_cover or tile_cover_sort:
+        tile_geom = polygons(mercantile.feature(tile)["geometry"]["coordinates"][0])
+        int_pcts = _intersect_percent(tile_geom, geoms)
+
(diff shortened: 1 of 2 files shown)


cogeo_mosaic/mosaic.py
```diff
@@ -14,7 +14,37 @@
 from pydantic import BaseModel, Field
 
 
-DEFAULT_ACCESSOR: Callable = lambda feature: feature["properties"]["path"]
+def default_accessor(feature: Dict):
```
vincentsarago, line 17: I've changed this to be consistent with `default_filter`

cogeo_mosaic/mosaic.py
```diff
@@ -97,13 +99,15 @@ def _create_mosaic(
         quadkey_zoom: int, optional
             Force mosaic quadkey zoom.
         accessor: callable, required
-            Function called on each feature to get its identifier (default is feature["properties"]["path"])
+            Function called on each feature to get its identifier (default is feature["properties"]["path"]).
+        asset_filter: callable, required
```
vincentsarago, line 103: This is definitely a really advanced feature because any `custom filter method` should match the inputs of the actual method (e.g `tile, intersect_dataset, intersect_geoms`) I'm fine with this but we will need documentation

kylebarron: Yes, docs required

tests/test_create.py
```diff
@@ -20,11 +21,16 @@
         mosaic_content["tiles"][qk] = [os.path.join(basepath, item) for item in asset]
 
 
+def _filter_and_sort(*args, **kwargs):
+    dataset = default_filter(*args, **kwargs)
+    return sorted(dataset, key=lambda x: x["properties"]["path"], reverse=True)
```
kylebarron, line 26: You need to use `accessor(x)` here.

kylebarron: Or sort within the filter

vincentsarago: I mean sure, but this is just for this test.

kylebarron: Oh sorry I didn't see this was a test

PR #70, opened 2020-05-01
add "quadkeys" properties and fix overviews for dynamoDB

vincentsarago:
This PR adds `.quadkeys` property for the backends in order to retrieve the list of quadkeys in the mosaicjson. 

For dynamoDB we are doing a scan operation but returning only the quadkey value (not the full item) so with relatively small mosaic the body should be <1Mb. We could add pagination if needed later

cc @kylebarron

### DIFF
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -60,6 +60,11 @@ class BaseBackend(AbstractContextManager):
         """Return Quadkey zoom property."""
         return self.mosaic_def.quadkey_zoom or self.mosaic_def.minzoom
 
+    @property
+    def quadkeys(self) -> List[str]:
+        """Return the list of quadkey tiles."""
+        return [qk for qk, _ in self.mosaic_def.tiles.items()]
+
     @abc.abstractmethod
     def write(self):
         """Upload new MosaicJSON to backend."""
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -61,9 +61,9 @@ class BaseBackend(AbstractContextManager):
         return self.mosaic_def.quadkey_zoom or self.mosaic_def.minzoom
 
     @property
-    def quadkeys(self) -> List[str]:
-        """Return the list of quadkey tiles."""
-        return [qk for qk, _ in self.mosaic_def.tiles.items()]
+    def _quadkeys(self) -> List[str]:
+        """Return the list of quadkeys"""
+        return list(self.mosaic_def.tiles.keys())
 
     @abc.abstractmethod
     def write(self):
--- a/cogeo_mosaic/backends/dynamodb.py
+++ b/cogeo_mosaic/backends/dynamodb.py
@@ -48,6 +48,12 @@ class DynamoDBBackend(BaseBackend):
         tile = mercantile.tile(lng, lat, self.quadkey_zoom)
         return self.get_assets(tile.x, tile.y, tile.z)
 
+    @property
+    def quadkeys(self) -> List[str]:
+        """Return the list of quadkey tiles."""
+        resp = self.table.scan(ProjectionExpression="quadkey")
+        return [qk["quadkey"] for qk in resp["Items"] if qk["quadkey"] != "-1"]
+
     def write(self):
         """Write mosaicjson document to AWS DynamoDB."""
         self._create_table()
--- a/cogeo_mosaic/backends/dynamodb.py
+++ b/cogeo_mosaic/backends/dynamodb.py
@@ -49,9 +49,12 @@ class DynamoDBBackend(BaseBackend):
         return self.get_assets(tile.x, tile.y, tile.z)
 
     @property
-    def quadkeys(self) -> List[str]:
+    def _quadkeys(self) -> List[str]:
         """Return the list of quadkey tiles."""
-        resp = self.table.scan(ProjectionExpression="quadkey")
+        warnings.warn(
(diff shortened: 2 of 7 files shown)


cogeo_mosaic/backends/dynamodb.py
```diff
@@ -48,6 +48,12 @@ def point(self, lng: float, lat: float) -> List[str]:
         tile = mercantile.tile(lng, lat, self.quadkey_zoom)
         return self.get_assets(tile.x, tile.y, tile.z)
 
+    @property
+    def quadkeys(self) -> List[str]:
+        """Return the list of quadkey tiles."""
+        resp = self.table.scan(ProjectionExpression="quadkey")
```
vincentsarago, line 54: Not sure how costly are the `scan` operation! maybe we could add a warning here ?

cogeo_mosaic/backends/base.py
```diff
@@ -60,6 +60,11 @@ def quadkey_zoom(self) -> int:
         """Return Quadkey zoom property."""
         return self.mosaic_def.quadkey_zoom or self.mosaic_def.minzoom
 
+    @property
+    def quadkeys(self) -> List[str]:
+        """Return the list of quadkey tiles."""
+        return [qk for qk, _ in self.mosaic_def.tiles.items()]
```
kylebarron, line 66: You can just do `list(self.mosaic_def.tiles.keys())`.

I think that's clearer than a list comprehension here

cogeo_mosaic/backends/dynamodb.py
kylebarron: Yeah I would suggest a warning.

I know that it's standard to return a `list`, but I'd suggest seeing how easy it is to return an iterator here.

kylebarron:
I also think we should be careful to not call `.quadkeys` internally, and leave that only to the user

vincentsarago:
> I also think we should be careful to not call .quadkeys internally, and leave that only to the user

I'm not sure to understand @kylebarron

kylebarron:
Since it's a costly function, presumably both in time and maybe in money, we shouldn't call `.quadkeys` from some other function. The only way it should be run is directly from the user, so that they know they're calling the scan if they're using dynamodb

vincentsarago:
@kylebarron well we need to call `.quadkeys` for the overview function. I think adding a note in the documentation that using Overview for DynamoDB will need scaning through the table

vincentsarago:
🤔 alright because I do not need all the quadkeys (we just need 1, to fetch data information) I think I'll add another `property` called `random_quakey` which will return only 1 quadkey value. We will still use the scan operation but with filters to return only one value

kylebarron:
I think that's fine because that's not something that would be called in a lambda tiler every request.

vincentsarago:
> 🤔 alright because I do not need all the quadkeys (we just need 1, to fetch data information) I think I'll add another property called random_quakey which will return only 1 quadkey value. We will still use the scan operation but with filters to return only one value

Well this is not True because I was also planning to use the list of quadkey to reduce the number of db call. I'll go forward with the documentation update and warning.

kylebarron:
Oh we posted comments at the same time.

> because I do not need all the quadkeys (we just need 1, to fetch data information)

I don't understand why this is. I also have looked it over a couple times and don't completely understand what the `overview` command does? It generates a _new_ COG at the higher zoom level? Or it adds overviews within an existing COG?

vincentsarago:
> I don't understand why this is. I also have looked it over a couple times and don't completely understand what the overview command does? It generates a new COG at the higher zoom level? Or it adds overviews within an existing COG?

It creates a lower zoom level version of the mosaic. e.g if you have a mosaic with minzoom at 10, `overview` will create a COG from the mosaic to cover zoom 3 to 9 (the minzoom is configurable). 

But to be able to create the COG I need to know:
- how many bands the data have
- what's the datatype
- what's the nodata value
- what's the mask/alpha info

The `overview` command is exactly what you needed for https://github.com/kylebarron/landsat-mosaic-latest/issues/6

kylebarron:
> The `overview` command is exactly what you needed for [kylebarron/landsat-mosaic-latest#6](https://github.com/kylebarron/landsat-mosaic-latest/issues/6)

Oh, cool! As you can tell I completely missed that.

> But to be able to create the COG I need to know:

Good points. So as long as the entire dataset is made up of homogeneous data, you only need to query metadata from a single source COG to find this info.


> We will still use the scan operation but with filters to return only one value

You can probably just do something like (pseudocode) `SELECT quadkey LIMIT 2;`, and then take one of those, choosing the one not `-1`.

vincentsarago:
> You can probably just do something like (pseudocode) SELECT quadkey LIMIT 2;, and then take one of those, choosing the one not -1.

yes that's what I though I was going to do:

>  I think I'll add another property called random_quakey which will return only 1 quadkey value. We will still use the scan operation but with filters to return only one value

But then
> Well this is not True because I was also planning to use the list of quadkey to reduce the number of db call. I'll go forward with the documentation update and warning.

I think having the full list of quadkeys is still nice so we reduce the number of calls to the dynamoDB table. Anyway with documentation and warning, users should understand that doing `overview` might result in some cost.

kylebarron:
Each individual record scanned is the same price as a `GET` on that record: https://stackoverflow.com/a/31540178.

I might suggest this to be `opt-in` for DynamoDB, kinda like how you have to append `--request-payer` to access an S3 bucket with requester pays enabled. I would think that a user will probably only use DynamoDB for large collections that are too large for S3.

PR #74, opened 2020-05-07
Add upload cli

kylebarron:
Adds a cli command to upload a prebuilt mosaic to a backend

### DIFF
--- a/cogeo_mosaic/scripts/cli.py
+++ b/cogeo_mosaic/scripts/cli.py
@@ -94,7 +94,9 @@ def create(
 
 @cogeo_cli.command(short_help="Upload mosaic definition to backend")
 @click.argument("file", type=click.File(mode="r"), default="-")
-@click.option("--url", type=str, help="URL to which the mosaic should be uploaded.")
+@click.option(
+    "--url", type=str, required=True, help="URL to which the mosaic should be uploaded."
+)
 def upload(
     file, url,
 ):
--- a/cogeo_mosaic/scripts/cli.py
+++ b/cogeo_mosaic/scripts/cli.py
@@ -92,6 +92,19 @@ def create(
         click.echo(json.dumps(mosaicjson.dict(exclude_none=True)))
 
 
+@cogeo_cli.command(short_help="Upload mosaic definition to backend")
+@click.argument("input", type=click.File(mode="r"), default="-")
+@click.option("--url", type=str, help="URL to which the mosaic should be uploaded.")
+def upload(
+    file, url,
+):
+    """Upload mosaic definition file."""
+    mosaicjson = json.loads(file)
+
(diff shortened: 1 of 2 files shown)


cogeo_mosaic/scripts/cli.py
```diff
@@ -92,6 +92,19 @@ def create(
         click.echo(json.dumps(mosaicjson.dict(exclude_none=True)))
 
 
+@cogeo_cli.command(short_help="Upload mosaic definition to backend")
+@click.argument("input", type=click.File(mode="r"), default="-")
+@click.option("--url", type=str, help="URL to which the mosaic should be uploaded.")
```
vincentsarago, line 97: this options should be `required`

cogeo_mosaic/scripts/cli.py
```diff
@@ -92,6 +92,19 @@ def create(
...
+def upload(
+    file, url,
+):
```
vincentsarago, line 100: does isort or black format this line like this ?

kylebarron: I ran black, but maybe it didn't propagate?

kylebarron: Actually it's because of the `,` after `url`

cogeo_mosaic/scripts/cli.py
kylebarron: Done

vincentsarago:
Open question: should we setup a test incase of existing table or existing file on s3 ?

kylebarron:
> Edit: DynamoDB will error if the table already exist

Will it? I don't think so. I think the default now is that it'll overwrite.

kylebarron:
It does error, but it's in a `try/except`:
https://github.com/developmentseed/cogeo-mosaic/blob/aa06bd11fabed99306e5b6205694267874af1ff0/cogeo_mosaic/backends/dynamodb.py#L125-L139

kylebarron:
But regardless, those test cases are unrelated to the CLI specifically

vincentsarago:
> But regardless, those test cases are unrelated to the CLI specifically

good point, I'll open an issue and see if we want to add `--force` option if the table or the file on s3 already exits

kylebarron:
But note that with dynamodb, the url must be specified with the region here:
```
dynamodb://us-west-2/table_name
```
the shorthand
```
dynamodb:///table_name
```
only works when the region is in an environment variable

vincentsarago:
> only works when the region is in an environment variable

I think it's fine, I don't really want to over complicate the CLI, DynamoDB is a more advanced backend. Maybe we can update https://github.com/developmentseed/cogeo-mosaic#mosaic-storage-backends to make sure people know about the default region

vincentsarago:
thanks @kylebarron this is great. I'll merge it in master but not sure if I will cut a release right away!

PR #78, opened 2020-05-25
Scaffolding for simple Mkdocs Material website

kylebarron:
- Adds `.travis.yml` file to auto-build/deploy website on new commits. You'd have to enable Travis CI for the repository and add a Github token as the `GITHUB_TOKEN` environment variable that allows Travis to push to the `gh-pages` branch. You probably want to also change the git config name/email.
- Adds a symlink for `docs/index.md` to the top-level README. Mkdocs only sees docs in the `docs/` folder.
- Adds a pretty standard `mkdocs.yml` config. You'll probably want to change `site_url` to whatever the public url would be.

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: python
+notifications:
+  email: false
+python:
+    - "3.7"
+script:
+  - git config user.name "Kyle Barron";
+  - git config user.email "kylebarron2@gmail.com";
+  - git remote add gh-token "https://developmentseed:${GITHUB_TOKEN}@github.com/developmentseed/cogeo-mosaic.git";
+  - git fetch gh-token && git fetch gh-token gh-pages:gh-pages;
+  - pip install -U mkdocs mkdocs-material pygments;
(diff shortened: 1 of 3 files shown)


vincentsarago:
@kylebarron I'm a bit concerned that we have to use travis. this repo is already set up with circleci (and I'd love to move to github actions). have 2 CI running might makes things a but complex. Is it mandatory or can build be setup in circleCI ?

kylebarron:
Not mandatory at all. I only included it because I'd already set up my own repos with it. You can

- Always deploy manually: it's literally just `mkdocs gh-deploy`. But obviously it doesn't auto-build and you have remember to re-deploy on every docs update.
- Figure out the steps to build with CircleCI/Github Actions.

	It looks like there's already an extension to do this with Github Actions: https://github.com/marketplace/actions/deploy-mkdocs. So it might be really easy, I just haven't tried.

vincentsarago:
closed in #114  
thanks for your help @kylebarron 🙏

Issue #81, opened 2020-06-21
Can't create mosaic when geometry have different number of points ? 


vincentsarago:
Bug found while working on #80 

```python
import requests
from pygeos import polygons
headers = {
    "Content-Type": "application/json",
    "Accept-Encoding": "gzip",
    "Accept": "application/geo+json",
}
data = requests.post("https://earth-search.aws.element84.com/v0/search", headers=headers).json()
dataset_geoms = polygons(
    [feat["geometry"]["coordinates"][0] for feat in data["features"]]
)
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-13-083d9d799f2d> in <module>
      1 dataset_geoms = polygons(
----> 2     [feat["geometry"]["coordinates"][0] for feat in data["features"]]
      3 )

pygeos/creation.py in polygons(shells, holes)
     88 
     89     if holes is None:
---> 90         return lib.polygons_without_holes(shells)
     91 
     92     holes = np.asarray(holes)

TypeError: One of the arguments is of incorrect type. Please provide only Geometry objects.
```

seems to be related to https://github.com/pygeos/pygeos/issues/138 

cc @kylebarron

vincentsarago:
Quick fix ? 

```python
try:
    dataset_geoms = polygons(
        [feat["geometry"]["coordinates"][0] for feat in features]
    )
except TypeError:
    dataset_geoms = [
        [polygons(feat["geometry"]["coordinates"][0]) for feat in features]
    ]
```

vincentsarago:
fixed in #82

PR #82, opened 2020-06-22
add stac backend

### LINKED ISSUE #80: Add STAC Backend ?
Questions:
- should we assume a STAC API + Query as input? 

```python
with STACBackend("https://jeffstacapi.hero", query=query, minzoom=4, maxzoom=12, quadkey=4) as mosaic:
```

On `init` the backend will `POST` to the api and construct the mosaic from the response 

Note: 
- This backend will be Read-Only (no write method)
- The POST request will be cached
- The List of asset will be composed of Item URLs

cc @geospatial-jeff

vincentsarago:
closes #80


cogeo_mosaic/mosaic.py
```diff
@@ -137,10 +137,17 @@ def _create_mosaic(
...
+        # Ref: https://github.com/developmentseed/cogeo-mosaic/issues/81
+        try:
+            dataset_geoms = polygons(
+                [feat["geometry"]["coordinates"][0] for feat in features]
+            )
+        except TypeError:
+            dataset_geoms = [
+                polygons(feat["geometry"]["coordinates"][0]) for feat in features
+            ]
+
```
vincentsarago, line 150: this fixes #81

cogeo_mosaic/backends/stac.py
```diff
@@ -0,0 +1,101 @@
...
+    features: List[Dict] = []
+
+    headers = {
+        "Content-Type": "application/json",
+        "Accept-Encoding": "gzip",
+        "Accept": "application/geo+json",
+    }
+
+    def _fetch(url):
+        return requests.post(url, headers=headers, data=query).json()
```
vincentsarago, line 83: Should we also support GET ?

cogeo_mosaic/backends/stac.py
```diff
@@ -0,0 +1,101 @@
...
+
+    next_url = stac_url
+    while True:
+        results = _fetch(next_url)
+        if results["context"]["returned"] == 0:
```
vincentsarago, line 88: is the context part of the spec?

cogeo_mosaic/backends/stac.py
```diff
@@ -0,0 +1,101 @@
...
+            break
+
+        features.extend(results["features"])
+        if max_items and len(features) >= max_items:
+            break
+
+        link = list(filter(lambda link: link["rel"] == "next", results["links"]))
```
vincentsarago, line 95: is `next` link part of the spec?

cogeo_mosaic/backends/stac.py
```diff
@@ -0,0 +1,101 @@
...
+import requests
+
+from cogeo_mosaic.backends.base import BaseBackend
+from cogeo_mosaic.backends.utils import get_assets_from_json
+from cogeo_mosaic.mosaic import MosaicJSON
+
+
+def default_stac_accessor(feature: Dict):
+    """Return specific feature identifier."""
+    link = list(filter(lambda link: link["rel"] == "self", feature["links"]))[0]
```
vincentsarago, line 17: is this part of the spec?

vincentsarago:
### ToDo
- [ ] check STAC API specification
- [ ] decide if we want to support stac in https://github.com/developmentseed/cogeo-mosaic/blob/master/cogeo_mosaic/backends/__init__.py#L12

☝️ this could be done with custom url like `stac+https://mystacapi.endpoint/search`

```python
def MosaicBackend(url: str, *args, **kwargs) -> BaseBackend:
    """Select mosaic backend for url."""
    parsed = urlparse(url)

    scheme = parsed.scheme.split("+")
    mosaic_scheme = scheme[0] or parsed.scheme

    if mosaic_scheme == "stac":
        url = url.replace("stac+", "")
        return STACBackend(url, *args, **kwargs)

    if mosaic_scheme == "s3":
        bucket = parsed.netloc
        key = parsed.path.strip("/")
        return S3Backend(bucket, key, *args, **kwargs)

    if mosaic_scheme == "dynamodb":
        table_name = parsed.path.strip("/")
        region = parsed.netloc
        if region:
            kwargs.update(dict(region=region))
        return DynamoDBBackend(table_name, *args, **kwargs)

    if mosaic_scheme in ["https", "http", "ftp"]:
        return HttpBackend(url, *args, **kwargs)

    if mosaic_scheme == "file":
        path = parsed.path
        return FileBackend(path, *args, **kwargs)

    return FileBackend(url, **kwargs)
```

cogeo_mosaic/backends/stac.py
vincentsarago: 🤔 https://github.com/radiantearth/stac-api-spec/blob/master/api-spec.md#paging-extension

vincentsarago: @geospatial-jeff any idea?

cogeo_mosaic/backends/stac.py
geospatial-jeff: Self link is strongly recommended but not required.  You could also build the item URL from `{base_url}/collections/{collectionID}/items/{itemID}` if self link is missing.

cogeo_mosaic/backends/stac.py
geospatial-jeff: There a few issues open on the api-spec repo to improve ambiguity such as this.  For now I think it would make sense to parameterize the rel type keeping `next` as the default.

cogeo_mosaic/backends/stac.py
geospatial-jeff: https://github.com/radiantearth/stac-api-spec/tree/master/extensions/context

cogeo_mosaic/backends/stac.py
geospatial-jeff: I don't think it is needed.  `POST /search` is [strongly recommended by the spec](https://github.com/radiantearth/stac-api-spec/blob/master/api-spec.md#post) and almost all implementations i'm aware of implement it because it has many advantages over GET

cogeo_mosaic/backends/stac.py
vincentsarago: I'll add something. 
Any way I think this MUST be covered by the documentation on the `accessor` option

cogeo_mosaic/backends/stac.py
```diff
@@ -0,0 +1,101 @@
...
+        """Fetch STAC API and construct the mosaicjson."""
+        features = stac_search(self.path, json.dumps(query), max_items=max_items)
+
+        return MosaicJSON.from_features(
+            features, minzoom, maxzoom, accessor=accessor, **kwargs
+        )
+
+
+@functools.lru_cache(maxsize=512)
+def stac_search(
```
vincentsarago, line 70: @kylebarron 👋 
ref: #83, if I out the decorator ontop of `_read` it won't be cached (meaning it will do the requests.post each time a new `STACBackend` is created). The solution was to create an `external` function that will be cached.

The downside is we are not caching the `Mosaic Creation` but only the HTTP requests

cogeo_mosaic/backends/stac.py
```diff
@@ -0,0 +1,151 @@
...
+        return os.path.join(
+            link[0]["href"],
+            "collections",
+            feature["collection"],
+            "items",
+            feature["id"],
+        )
+
+    # Fall back to the item ID
+    return feature["id"]
```
vincentsarago, line 33: @geospatial-jeff sadly this is ran at feature level so there is no real info about the stac api url. Here I choose to fallback first using the `root` and worse case to the `id` only. 

Again this will need documentation

cogeo_mosaic/backends/stac.py
```diff
@@ -0,0 +1,189 @@
...
+
+    next_url = stac_url
+    page = 1
+    while True:
+        # HACK: next token is not yet handled in some API
+        parsed_url = urlparse(next_url)
+        qs: Dict[str, Any] = dict(parse_qsl(parsed_url.query))
+        qs.update({"page": page})
+        updated_query = urlencode(qs)
+        next_url = parsed_url._replace(query=updated_query).geturl()
```
vincentsarago, line 165: This is a HACK to overcome some issues in current stac-server API

cogeo_mosaic/backends/stac.py
```diff
@@ -0,0 +1,189 @@
...
+        # Check if there is more data to fetch
+        if results["context"]["matched"] <= results["context"]["returned"]:
+            break
+
+        # Right now we construct the next URL
+        # https://github.com/radiantearth/stac-api-spec/blob/master/api-spec.md#paging-extension
+        # link = list(filter(lambda link: link["rel"] == next_link_key, results["links"]))
+        # if not link:
+        #     break
+        # next_url = link[0]["href"]
```
vincentsarago, line 185: The API is still to be fixed for the `next` page

PR #84, opened 2020-06-25
switch to cachetool cache

### LINKED ISSUE #83: LRU cache doesn't work on object method!
https://github.com/developmentseed/cogeo-mosaic/blob/7b6746a41ec8c5a2a9c618a9f9595ac3eccd85f9/cogeo_mosaic/backends/s3.py#L64

vincentsarago:
closes #83 
cc @kylebarron

### DIFF
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -17,7 +17,7 @@ class BaseBackend(AbstractContextManager):
     path: str
     mosaic_def: MosaicJSON
     _backend_name: str
-    _file_byte_size: Optional[int]
+    _file_byte_size: Optional[int] = 0
 
     @abc.abstractmethod
     def __init__(self, *args, **kwargs):
--- a/cogeo_mosaic/backends/dynamodb.py
+++ b/cogeo_mosaic/backends/dynamodb.py
@@ -1,6 +1,5 @@
 """cogeo-mosaic AWS DynamoDB backend."""
 
-import functools
 import itertools
 import json
 import os
@@ -12,6 +11,8 @@ from typing import Any, Dict, List, Optional, Sequence, Union
 import boto3
 import click
 import mercantile
+from cachetools import TTLCache, cached
+from cachetools.keys import hashkey
 
 from cogeo_mosaic.backends.base import BaseBackend
 from cogeo_mosaic.backends.utils import find_quadkeys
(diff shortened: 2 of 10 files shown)


cogeo_mosaic/backends/base.py
```diff
@@ -17,7 +17,7 @@ class BaseBackend(AbstractContextManager):
     path: str
     mosaic_def: MosaicJSON
     _backend_name: str
-    _file_byte_size: Optional[int]
+    _file_byte_size: Optional[int] = 0
```
vincentsarago, line 20: We have to set a default value here because when the backend hit the cache we cannot set the size.

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -165,7 +166,9 @@ def _write_items(self, items: List[Dict]):
                 for item in progitems:
                     batch.put_item(item)
 
-    @functools.lru_cache(maxsize=512)
+    @cached(
+        TTLCache(maxsize=512, ttl=300), key=lambda self: hashkey(self.path),
+    )
```
vincentsarago, line 171: simple cache, based only on the table + region name

cogeo_mosaic/backends/s3.py
```diff
@@ -61,7 +62,10 @@ def write(self, gzip: bool = None, **kwargs: Any):
 
         _aws_put_data(self.key, self.bucket, body, client=self.client, **kwargs)
 
-    @functools.lru_cache(maxsize=512)
+    @cached(
+        TTLCache(maxsize=512, ttl=300),
+        key=lambda self, gzip=None: hashkey(self.key, self.bucket, gzip),
```
vincentsarago, line 67: we could use `self.path`here instead of the key+bucket

cogeo_mosaic/backends/stac.py
```diff
@@ -130,10 +131,13 @@ def _read(  # type: ignore
         )
 
 
-@functools.lru_cache(maxsize=512)
+@cached(
+    TTLCache(maxsize=512, ttl=300),
+    key=lambda url, query, **kwargs: hashkey(url, json.dumps(query), **kwargs),
```
vincentsarago, line 136: query is not hashable to we jsonify it

tests/test_cli.py
```diff
@@ -59,31 +59,31 @@ def test_update_valid():
     """Should work as expected."""
     runner = CliRunner()
     with runner.isolated_filesystem():
-        with open("mosaic.json", "w") as f:
+        with open("mosaic_1.json", "w") as f:
```
vincentsarago, line 62: renamed to avoid caching

cogeo_mosaic/backends/dynamodb.py
kylebarron: Is `maxsize` in bytes or items?

vincentsarago: Good question, I thought it was in bytes, but it's in items! which is still ok I guess 💁

vincentsarago: > All these decorators wrap a function with a memoizing callable that saves up to the maxsize most recent calls, u

PR #90, opened 2020-07-11
Default to 1 thread for Mosaic overview creation

kylebarron:
I finally have the most conclusive evidence yet of GDAL(?) issues with threading.

```bash
export AWS_REQUEST_PAYER="requester"
cogeo-mosaic overview naip_2016_2018_032023.json
mv naip_2016_2018_032023_0.tif naip_2016_2018_032023_0.bak.tif
cogeo-mosaic overview --threads=1 naip_2016_2018_032023.json
```

[naip_2016_2018_032023.json.zip](https://github.com/developmentseed/cogeo-mosaic/files/4906272/naip_2016_2018_032023.json.zip)

Since my computer has 8 physical CPU cores and 16 logical cores, the default setting of `multiprocessing.cpu_count() * 5` uses **_80_** threads.

Default (80) threads: 
image

One thread: (note that the remaining black areas are expected; NAIP images aren't taken over water)
image

### DIFF
--- a/cogeo_mosaic/scripts/cli.py
+++ b/cogeo_mosaic/scripts/cli.py
@@ -268,8 +268,8 @@ def footprint(input_files, output, threads, quiet):
 @click.option(
     "--threads",
     type=int,
-    default=lambda: os.environ.get("MAX_THREADS", multiprocessing.cpu_count() * 5),
-    help="threads",
+    default=1,
+    help="# of Python threads. 1 thread is recommended, as GDAL can have issues when multiple threads read the same file.",
 )
 @click.option(
     "--overview-level",


kylebarron:
Of note: I'd previously successfully made an overview of NAIP in Iowa using many threads, but I think the key here is the shuffle:

https://github.com/developmentseed/cogeo-mosaic/blob/9986e4f743f6476160267a43b4c1e9ab65d7b7f8/cogeo_mosaic/overviews.py#L116-L118

With a sufficiently large mosaic, that shuffle means that it's unlikely that any threads will share the same source tiffs.

In order to keep threading on by default, I'd argue we need some way to make sure GDAL is never accessing the same source file at once in multiple threads. Note that this is complex because multiple quadkeys can reference the same source image.

image

PR #91, opened 2020-07-11
Small overview refactor

kylebarron:
These changes are for discussion. I'm using this code right now to create NAIP overviews

- Adds low-memory mode, where fetched images are written directly to a GeoTIFF on disk.
- Uses `dict` instead of `tuple` for `info` internally, for clearer code

### DIFF
--- a/cogeo_mosaic/overviews.py
+++ b/cogeo_mosaic/overviews.py
@@ -45,10 +45,8 @@ def _split_extrema(extrema: Dict, max_ovr: int = 6, tilesize: int = 256):
     """Create multiple extremas."""
     nb_ovrtile = 2 ** max_ovr
     extr = []
-    for _, x in enumerate(range(extrema["x"]["min"], extrema["x"]["max"], nb_ovrtile)):
-        for _, y in enumerate(
-            range(extrema["y"]["min"], extrema["y"]["max"], nb_ovrtile)
-        ):
+    for x in range(extrema["x"]["min"], extrema["x"]["max"], nb_ovrtile):
+        for y in range(extrema["y"]["min"], extrema["y"]["max"], nb_ovrtile):
             maxx = (
                 x + nb_ovrtile
                 if x + nb_ovrtile < extrema["x"]["max"]
@@ -65,8 +63,8 @@ def _split_extrema(extrema: Dict, max_ovr: int = 6, tilesize: int = 256):
 
 def _get_blocks(extrema):
     for j in range(extrema["y"]["max"] - extrema["y"]["min"]):
-        row = j * 256
         for i in range(extrema["x"]["max"] - extrema["x"]["min"]):
+            row = j * 256
             col = i * 256
             yield (j, i), Window(col_off=col, row_off=row, width=256, height=256)
 
--- a/cogeo_mosaic/overviews.py
+++ b/cogeo_mosaic/overviews.py
@@ -171,7 +171,7 @@ def create_low_level_cogs(
                                 length=len(future_work),
                                 show_percent=True,
                             ) as future:
-                                for res in future:
+                                for result in future:
                                     pass
 
                         for f in _filter_futures(future_work):
--- a/cogeo_mosaic/overviews.py
+++ b/cogeo_mosaic/overviews.py
@@ -47,16 +47,8 @@ def _split_extrema(extrema: Dict, max_ovr: int = 6, tilesize: int = 256):
     extr = []
     for x in range(extrema["x"]["min"], extrema["x"]["max"], nb_ovrtile):
         for y in range(extrema["y"]["min"], extrema["y"]["max"], nb_ovrtile):
-            maxx = (
-                x + nb_ovrtile
-                if x + nb_ovrtile < extrema["x"]["max"]
-                else extrema["x"]["max"]
-            )
-            maxy = (
-                y + nb_ovrtile
-                if y + nb_ovrtile < extrema["y"]["max"]
-                else extrema["y"]["max"]
-            )
+            maxx = min(x + nb_ovrtile, extrema["x"]["max"])
+            maxy = min(y + nb_ovrtile, extrema["y"]["max"])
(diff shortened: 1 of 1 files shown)


cogeo_mosaic/overviews.py
```diff
@@ -30,54 +30,59 @@ def _get_info(asset: str) -> Tuple:
...
+    """Open rasterio object for writing overview image
+    """
+    if in_memory:
+        with MemoryFile() as memfile:
+            return memfile.open(**kwargs)
+
+    if not path:
+        raise ValueError("path must be provided when in_memory is False")
+
+    return rasterio.open(path, "w", **kwargs)
```
kylebarron, line 74: When used from inside `rasterio_open`, does it still work as a context manager, removing finished resources?

kylebarron: Maybe I need to add `@contextlib.contextmanager`...

cogeo_mosaic/overviews.py
```diff
@@ -30,54 +31,59 @@ def _get_info(asset: str) -> Tuple:
...
-        )
+        return {
+            "n_bands": src_dst.count,
+            "dtype": src_dst.dtypes[0],
+            "band_desc": description,
+            "tags": src_dst.tags(),
+            "nodata": src_dst.nodata,
+            "mask": mask,
+            "indexes": indexes,
+        }
```
vincentsarago, line 42: 👍

cogeo_mosaic/overviews.py
```diff
@@ -30,54 +31,59 @@ def _get_info(asset: str) -> Tuple:
...
-            )
-            maxy = (
-                y + nb_ovrtile
-                if y + nb_ovrtile < extrema["y"]["max"]
-                else extrema["y"]["max"]
-            )
+    for x in range(extrema["x"]["min"], extrema["x"]["max"], nb_ovrtile):
+        for y in range(extrema["y"]["min"], extrema["y"]["max"], nb_ovrtile):
+            maxx = min(x + nb_ovrtile, extrema["x"]["max"])
+            maxy = min(y + nb_ovrtile, extrema["y"]["max"])
```
vincentsarago, line 52: 👍

cogeo_mosaic/overviews.py
vincentsarago: maybe we could do the same as in rio-cogeo https://github.com/cogeotiff/rio-cogeo/blob/master/rio_cogeo/cogeo.py#L214-L221

kylebarron: In `rio-cogeo` you still write to a `TemporaryFile`... is that necessary or can you write directly to the output GeoTIFF?

One question is whether the output GeoTIFF _needs_ to have overviews. E.g. if we could do 
```
cogeo-mosaic overview ... out.tif
rio cogeo create out.tif out_cog.tif
```
then you wouldn't have to copy so much from `rio-cogeo`. Right now you copy most of the CLI arguments from `rio-cogeo` here, but I think the overview code would be a lot simpler by keeping the two parts separate.

kylebarron:
One of the main points of this refactor was to try to keep memory usage low by writing to file. As mentioned on Slack, I can tell that the output file has 0 bytes until the process has finished. Therefore this is still holding the entire raster in memory at some point.

I'd love to fix this so that each window is written to disk and then no longer held in memory, but I don't know how to do that.

cogeo_mosaic/overviews.py
```diff
@@ -124,80 +128,67 @@ def create_low_level_cogs(
...
 
-                            mem.write(tile, window=window)
-                            if info[5]:
-                                mem.write_mask(mask.astype("uint8"), window=window)
+                            dst.write(tile, window=window)
+                            if info["mask"]:
+                                dst.write_mask(mask.astype("uint8"), window=window)
 
+                    # Not already written to a file
+                    if in_memory:
```
vincentsarago, line 187: so if you don't use `in_memory` you don't create a COG? 👎

kylebarron: https://github.com/developmentseed/cogeo-mosaic/pull/91#discussion_r453224648

vincentsarago: well you can't create a COG directly you need to perform a `copy` (which is what rio-cogeo does)

cogeo_mosaic/overviews.py
```diff
@@ -124,80 +128,67 @@ def create_low_level_cogs(
...
+                            cogeo.tile,
+                            indexes=info["indexes"],
+                            tilesize=tilesize,
+                            pixel_selection=defaults.FirstMethod(),
+                        )
+                        return window, tile, mask
+
+                    with futures.ThreadPoolExecutor(max_workers=threads) as executor:
+                        future_work = executor.map(_get_tile, blocks)
+                        for result in tqdm(future_work, total=len(blocks)):
```
vincentsarago, line 177: 👋 any reason to use tqdm instead of click progressbar ?

kylebarron: I couldn't get it to show with `click.progressbar`. Maybe it was something with `futures` that I just didn't understand, but I had to switch to `tqdm` to get some sort of progress bar to show

cogeo_mosaic/overviews.py
kylebarron: It would be cool if somehow you could import the cli options and then forward them to `rio-cogeo`. Something like
```py
import rio_cogeo

@click.command()
@rio_cogeo.click_options()
def overview(**rio_cogeo_options):
	...
```
And then not need to copy them here. But I'm not sure how that would work

cogeo_mosaic/overviews.py
kylebarron: I understand; ideally it would be nice to support a low-memory option, where you first write a geotiff using windows and then on the second pass you call `rio-cogeo` on that geotiff with a low-memory option there too.

vincentsarago:
this has been taken care of in https://github.com/developmentseed/cogeo-mosaic/pull/95

PR #94, opened 2020-07-28
add cache configuration

### LINKED ISSUE #93: Configurable cache
I think the following `cachetools` settings should be configurable through environment variables:
- cache `maxsize`
- cache `ttl`

It would also be nice to optionally disable the cogeo-mosaic cache for apps which implement their own higher-level caching strategies (ex. titiler).  This could be achieved by wrapping the `cachetools` cache in a decorator which optionally caches the result of the function being decorated based on an environment variable.

geospatial-jeff:
Implementation is based on [`cache-tools`](https://github.com/tkem/cachetools/blob/master/cachetools/decorators.py#L6-L44)

Closes #93

### DIFF
--- a/cogeo_mosaic/cache.py
+++ b/cogeo_mosaic/cache.py
@@ -1,0 +1,39 @@
+"""backend cache configuration"""
+import functools
+import os
+from typing import Callable
+
+from cachetools import TTLCache, cached
+
+# TTL of the cache in seconds
+CACHE_TTL = int(os.getenv("CACHE_TTL", 300))
+
+# Maximum size of the LRU cache in MB
+CACHE_MAXSIZE = int(os.getenv("CACHE_MAXSIZE", 512))
+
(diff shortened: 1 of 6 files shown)


cogeo_mosaic/cache.py
```diff
@@ -0,0 +1,39 @@
...
+import os
+from typing import Callable
+
+from cachetools import TTLCache, cached
+
+# TTL of the cache in seconds
+CACHE_TTL = int(os.getenv("CACHE_TTL", 300))
+
+# Maximum size of the LRU cache in MB
+CACHE_MAXSIZE = int(os.getenv("CACHE_MAXSIZE", 512))
```
geospatial-jeff, line 12: Maybe these env variables should be prefixed with `COGEO_MOSAIC`

vincentsarago: 👍

PR #95, opened 2020-07-29
allow pixel_selection and --not-in-memory option for overview cli

### LINKED ISSUE #89: Rewriting value of `res`
Here you're rewriting the value of `res`:
https://github.com/developmentseed/cogeo-mosaic/blob/9986e4f743f6476160267a43b4c1e9ab65d7b7f8/cogeo_mosaic/overviews.py#L185

So in the second iteration of the loop, you get:
```
TypeError: bad operand type for unary -: 'Future'
```

vincentsarago:
closes #89 

also do:
- update for latest rio-tiler
- add `--in-memory/--no-in-memory` and `--method` to the overview command

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,10 @@
+3.0a6 (TBD)
+------------------
+- Use environement variable to set/disable cache (#93, autho @geospatial-jeff)
+- Allow Threads configuration for overview command (author @kylebarron)
+- add --in-memory/--no-in-memory to control temporary files creation for `overview` function. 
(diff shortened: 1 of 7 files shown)


kylebarron:
At some point I'll try out the `--no-in-memory` and see how much of an impact that makes.

vincentsarago:
> At some point I'll try out the --no-in-memory and see how much of an impact that makes.

@kylebarron yes just make sure to set GDAL_CACHEMAX low or GDAL will still fill the memory

Issue #98, opened 2020-07-31
value -> values? 


vincentsarago:
https://github.com/developmentseed/cogeo-mosaic/blob/c989249ecda0ab090e5fff86e6aba2ba8c882f3f/cogeo_mosaic/backends/base.py#L84

PR #99, opened 2020-08-01
adapt for rio-tiler 2.0b5

### LINKED ISSUE #97: Add specific tests for tile and point
I forgot the write tests for both tile and point 🤦 

https://github.com/developmentseed/cogeo-mosaic/blob/c989249ecda0ab090e5fff86e6aba2ba8c882f3f/cogeo_mosaic/backends/base.py#L59-L97

vincentsarago:
also closes #97 and #98

### DIFF
--- a/cogeo_mosaic/backends/__init__.py
+++ b/cogeo_mosaic/backends/__init__.py
@@ -15,7 +15,7 @@ def MosaicBackend(url: str, *args: Any, **kwargs: Any) -> BaseBackend:
     """Select mosaic backend for url."""
     parsed = urlparse(url)
 
-    if parsed.scheme.startswith("stac+"):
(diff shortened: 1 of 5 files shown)


cogeo_mosaic/backends/__init__.py
```diff
@@ -15,7 +15,7 @@ def MosaicBackend(url: str, *args: Any, **kwargs: Any) -> BaseBackend:
     """Select mosaic backend for url."""
     parsed = urlparse(url)
 
-    if parsed.scheme.startswith("stac+"):
+    if parsed.scheme and parsed.scheme.startswith("stac+"):
```
vincentsarago, line 18: found a bug ;-) (tests added)

PR #111, opened 2020-08-30
WIP rename 'asset' (meaning STAC Item) to 'Item'

kylebarron:


vincentsarago:
Just adding some info here.

In stac we use Assets to define Items assets and thus we use this term in all other project to define the assets in each STAC items (also to define bands file in rio-tiler-pds). 
For mosaics, if we follow the STAC naming convention the word Items is what we should use here instead of assets.

This is a pretty big change but it will avoid some confusion for the users

kylebarron:
Closing as the core issue (the term `asset` referencing two different ideas) was fixed in https://github.com/cogeotiff/rio-tiler-pds/pull/13, which renamed `asset` to `band` in `rio-tiler-pds`

PR #123, opened 2020-10-15
Add sub-table name for dynamodb storage

### LINKED ISSUE #116: Multiple mosaics within a single DynamoDB table
AWS has a limit of 256 DynamoDB tables per region per account. Since DynamoDB is the most efficient mosaic storage type for very large mosaics, it would be nice to be able to store multiple mosaics in a single table.

According to @geospatial-jeff and [this AWS blog post](https://aws.amazon.com/blogs/database/choosing-the-right-dynamodb-partition-key/), mosaic keys of the sort `${mosaic-name}-${quadkey}` would maintain efficient queries.

Along the same lines: how to store top-level mosaicjson metadata. Currently the metadata for a table is stored as quadkey `-1`. It would be straightforward to keep metadata files as `${mosaic-name}--1` (an extra `-` for `-1`).

Another good question is how to keep track of all the mosaics in a single table. I'm guessing that listing all the keys that are like `.*-1` is inefficent. Would it be worthwhile to have another file that lists the key names within the table?

The parameter could be named something like `table_subname`, and default to `None` for backwards compatibility.

@geospatial-jeff

kylebarron:
Closes #116 

- What should be the name of the subkey? `subtable_name` doesn't sound great
- This defines the canonical path as `dynamodb://{region}/{table_name}:{subtable_name}`. A colon is not a valid DynamoDB table identifier
- I don't know why Mypy is unhappy with `cogeo_mosaic/backends/dynamodb.py:64: error: Module has no attribute "matches_re"`

It would probably also be good to have

- a `delete` method, so that you can delete one of the subtables. This would read the metadata for a single subkey and then iterate over its quadkeys, calling a delete operation on each one
- a key to list all subkeys that exist in the table. Use something like `-2` (?) (since we already use `-1` for the mosaic metadata), and have it contain a list of strings that are the subkeys that exist in the table. This would need to stay updated with the CRUD operations.
- Would we want to prevent a full overwrite of the table? Currently one of the methods will delete and create a new table if `overwrite=True`.

### DIFF
--- a/cogeo_mosaic/backends/dynamodb.py
+++ b/cogeo_mosaic/backends/dynamodb.py
@@ -23,22 +23,67 @@ from cogeo_mosaic.errors import _HTTP_EXCEPTIONS, MosaicError, MosaicExists
 from cogeo_mosaic.mosaic import MosaicJSON
 from cogeo_mosaic.utils import bbox_union
 
+# From https://docs.aws.amazon.com/general/latest/gr/rande.html
+AWS_REGIONS = [
+    "af-south-1",
+    "ap-east-1",
+    "ap-northeast-1",
+    "ap-northeast-2",
+    "ap-northeast-3",
+    "ap-south-1",
+    "ap-southeast-1",
+    "ap-southeast-2",
+    "ca-central-1",
+    "cn-north-1",
+    "cn-northwest-1",
+    "eu-central-1",
+    "eu-north-1",
+    "eu-south-1",
+    "eu-west-1",
+    "eu-west-2",
+    "eu-west-3",
+    "me-south-1",
+    "sa-east-1",
+    "us-east-1",
+    "us-east-2",
+    "us-west-1",
+    "us-west-2",
+]
+
 
 @attr.s
 class DynamoDBBackend(BaseBackend):
     """DynamoDB Backend Adapter."""
 
     client: Any = attr.ib(default=None)
-    region: str = attr.ib(default=os.getenv("AWS_REGION", "us-east-1"))
-    table_name: str = attr.ib(init=False)
+    region: str = attr.ib(
+        default=os.getenv("AWS_REGION", "us-east-1"),
+        validator=attr.validators.in_(AWS_REGIONS),
+    )
+    table_name: str = attr.ib(
+        init=False, validator=attr.validators.matches_re(r"^[a-zA-Z0-9\_\-\.]+$")
+    )
+    subtable_name: str = attr.ib(default=None)
     table: Any = attr.ib(init=False)
 
     _backend_name = "AWS DynamoDB"
 
     def __attrs_post_init__(self):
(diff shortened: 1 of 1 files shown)


cogeo_mosaic/backends/dynamodb.py
```diff
@@ -23,22 +23,67 @@
...
+        """
         parsed = urlparse(self.path)
-        self.table_name = parsed.path.strip("/")
+        table_id = parsed.path.lstrip("/").split(":")
+
+        self.table_name = table_id[0]
+        if len(table_id) > 1:
+            self.subtable_name = table_id[1]
+
         self.region = parsed.netloc or self.region
```
vincentsarago, line 87: what do you think about making region mandatory?

kylebarron: I think it would be fine to make that explicit.

vincentsarago:
@kylebarron thanks for starting this! 

With help from @leothomas I've created a new PR to handler this https://github.com/developmentseed/cogeo-mosaic/pull/127

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -23,22 +23,67 @@
...
 
     client: Any = attr.ib(default=None)
-    region: str = attr.ib(default=os.getenv("AWS_REGION", "us-east-1"))
-    table_name: str = attr.ib(init=False)
+    region: str = attr.ib(
+        default=os.getenv("AWS_REGION", "us-east-1"),
+        validator=attr.validators.in_(AWS_REGIONS),
+    )
+    table_name: str = attr.ib(
+        init=False, validator=attr.validators.matches_re(r"^[a-zA-Z0-9\_\-\.]+$")
```
vincentsarago, line 64: @kylebarron validators only run on init, so this won't work

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -23,22 +23,67 @@
...
 @attr.s
 class DynamoDBBackend(BaseBackend):
     """DynamoDB Backend Adapter."""
 
     client: Any = attr.ib(default=None)
-    region: str = attr.ib(default=os.getenv("AWS_REGION", "us-east-1"))
-    table_name: str = attr.ib(init=False)
+    region: str = attr.ib(
+        default=os.getenv("AWS_REGION", "us-east-1"),
+        validator=attr.validators.in_(AWS_REGIONS),
```
vincentsarago, line 61: I prefer not having a list to validate to because everytime AWS will add a new region we will need a new release

cogeo_mosaic/backends/dynamodb.py
vincentsarago: @kylebarron because the `table_name` is set during the `__post_init` the validation is not happening! you'll need to re-run the validation at the end of the `__post_init__`

PR #125, opened 2020-10-15
fix STAC pagination (E84 api)

### LINKED ISSUE #121: update stac pagination logic to match spec
stac spec defines some links keys which let the client formulate the correct request to fetch next/previous page.  This would allow for more robust pagination (get vs post etc.).

```
curl -X POST http://stac.arturo.ai/search --data '{"page": 3}' | jq '.links'
[
  {
    "href": "http://stac.arturo.ai/search",
    "rel": "next",
    "type": "application/geo+json",
    "title": null,
    "label:assets": null,
    "method": "POST",
    "body": {
      "token": "NmGhOLfn"
    },
    "merge": true
  }
]
```

vincentsarago:
closes #121

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,24 @@
 # Release Notes
 
+## Next (TBD)
+
+* add logger (`cogeo_mosaic.logger.logger`)
+* Update STACBackend to better handler paggination (ref: https://github.com/developmentseed/cogeo-mosaic/pull/125)
+* with change from #125, `stac_next_link_key` has be specified if you know the STAC API is using the latest specs:
+
+```python
+with MosaicBackend(
+    f"stac+{stac_endpoint}",
+    query.copy(),
+    11,
+    14,
+    backend_options={
+        "accessor": lambda feature: feature["id"],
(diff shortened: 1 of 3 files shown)


vincentsarago:
## TODO 
- [x] check with https://earth-search.aws.element84.com/v0/search
- [ ] check with sat-api.devseed
- [ ] check with arturo cc @geospatial-jeff 

other API we can test ?

vincentsarago:
FYI This is not working for your API @geospatial-jeff because you need the `next token` I think

vincentsarago:
The proposed script will work with old and new API. 

By default we will add `page=X` in the stac query (this is for old style API), For new API using `Next` links, user will need to use set the `stac_next_link_key` option to tell the backend to get the new query from the links.

```python
with MosaicBackend(
    f"stac+{stac_endpoint}",
    query.copy(),
    11,
    14,
    backend_options={
        "accessor": lambda feature: feature["id"],
        "stac_next_link_key": "next",
    }
) as mosaic:
``` 

cc @geospatial-jeff

Issue #126, opened 2020-10-16
Do not return the list of quadkey by default


vincentsarago:
https://github.com/developmentseed/cogeo-mosaic/blob/4946b9cb33290d1d90fcce7fdfffed6f69177936/cogeo_mosaic/backends/base.py#L77

This works well on small mosaic, but if you have 10000 quakdey (first mosaicjson file will be quite big) it won't be really efficient.

kylebarron:
For all backends?

vincentsarago:
@kylebarron yes, basically we should add an options like we do in dynamoDB https://github.com/developmentseed/cogeo-mosaic/blob/master/cogeo_mosaic/backends/dynamodb.py#L56

PR #127, opened 2020-10-16
Support Multi Mosaic in DynamoDB Table

### LINKED ISSUE #116: Multiple mosaics within a single DynamoDB table

vincentsarago:
closes #116

### DIFF
--- a/cogeo_mosaic/backends/dynamodb.py
+++ b/cogeo_mosaic/backends/dynamodb.py
@@ -317,7 +317,7 @@ class DynamoDBBackend(BaseBackend):
         return bool(item)
 
     def delete(self):
-        """Delete every items for a specific mosaic in the dynamoDB Table."""
+        """Delete all items for a specific mosaic in the dynamoDB Table."""
         logger.debug(f"Deleting all items for mosaic {self.mosaic_name}...")
 
         quadkey_list = self._quadkeys + [self._metadata_quadkey]
--- a/cogeo_mosaic/backends/dynamodb.py
+++ b/cogeo_mosaic/backends/dynamodb.py
@@ -52,7 +52,7 @@ class DynamoDBBackend(BaseBackend):
         logger.debug(f"Using DynamoDB backend: {self.path}")
 
         if not re.match(
-            r"dynamodb://([a-z]{2}\-[a-z]+\-[0-9])?\/[a-zA-Z0-9\_\-\.]+\:[a-zA-Z0-9\_\-\.]+$",
+            r"^dynamodb://([a-z]{2}\-[a-z]+\-[0-9])?\/[a-zA-Z0-9\_\-\.]+\:[a-zA-Z0-9\_\-\.]+$",
             self.path,
         ):
             raise ValueError(f"Invalid DynamoDB path: {self.path}")
--- a/cogeo_mosaic/backends/dynamodb.py
+++ b/cogeo_mosaic/backends/dynamodb.py
@@ -25,7 +25,7 @@ from cogeo_mosaic.mosaic import MosaicJSON
 from cogeo_mosaic.utils import bbox_union
 
 
-@attr.s()
+@attr.s
 class DynamoDBBackend(BaseBackend):
     """DynamoDB Backend Adapter."""
 
--- a/cogeo_mosaic/backends/dynamodb.py
+++ b/cogeo_mosaic/backends/dynamodb.py
@@ -294,4 +294,28 @@ class DynamoDBBackend(BaseBackend):
     def clean(self):
         """clean MosaicID from dynamoDB Table."""
         logger.debug(f"Deleting items for mosaic {self.mosaic_name}...")
-        pass
+
+        # get items
+        resp = self.table.query(
+            KeyConditionExpression="#mosaicId = :mosaicId",
+            ExpressionAttributeNames={"#mosaicId": "mosaicId", "#quadKey": "quadKey"},
+            ExpressionAttributeValues={":mosaicId": {"S": self.mosaic_name}},
+            ProjectionExpression="#quadKey",
+        )
+
+        # delete items
+        for i in resp["Items"]:
+            self.client.batch_write_item(
+                RequestItems={
+                    self.table_name: [
+                        {
+                            "DeleteRequest": {
+                                "Key": {
+                                    "mosaicId": {"S": i["mosaicId"]},
+                                    "quadkey": {"S": i["quadkey"]},
+                                }
+                            }
+                        }
+                    ]
+                }
+            )
--- a/cogeo_mosaic/backends/dynamodb.py
+++ b/cogeo_mosaic/backends/dynamodb.py
@@ -14,8 +14,10 @@ import attr
 import boto3
 import click
 import mercantile
+from boto3.dynamodb.conditions import Key
 from botocore.exceptions import ClientError
 from cachetools.keys import hashkey
+from rio_tiler.utils import _chunks
 
 from cogeo_mosaic.backends.base import BaseBackend
 from cogeo_mosaic.backends.utils import find_quadkeys
@@ -93,11 +95,8 @@ class DynamoDBBackend(BaseBackend):
     def _quadkeys(self) -> List[str]:
         """Return the list of quadkey tiles."""
         resp = self.table.query(
-            KeyConditionExpression="#mosaicId = :mosaicId",
-            # allows you to use dyanmodb reserved keywords as field names
-            ExpressionAttributeNames={"#mosaicId": "mosaicId", "#quadKey": "quadKey"},
-            ExpressionAttributeValues={":mosaicId": {"S": self.mosaic_name}},
-            ProjectionExpression="#quadKey",
+            KeyConditionExpression=Key("mosaicId").eq(self.mosaic_name),
+            ProjectionExpression="quadkey",
         )
         return [qk["quadkey"] for qk in resp["Items"] if qk["quadkey"] != "-1"]
 
@@ -296,27 +295,25 @@ class DynamoDBBackend(BaseBackend):
         """clean MosaicID from dynamoDB Table."""
         logger.debug(f"Deleting items for mosaic {self.mosaic_name}...")
 
-        # get items
+        # get quadkeys
         resp = self.table.query(
-            KeyConditionExpression="#mosaicId = :mosaicId",
-            ExpressionAttributeNames={"#mosaicId": "mosaicId", "#quadKey": "quadKey"},
-            ExpressionAttributeValues={":mosaicId": {"S": self.mosaic_name}},
-            ProjectionExpression="#quadKey",
+            KeyConditionExpression=Key("mosaicId").eq(self.mosaic_name),
+            ProjectionExpression="quadkey",
         )
 
         # delete items
-        for i in resp["Items"]:
-            self.client.batch_write_item(
-                RequestItems={
-                    self.table_name: [
-                        {
-                            "DeleteRequest": {
-                                "Key": {
-                                    "mosaicId": {"S": i["mosaicId"]},
-                                    "quadkey": {"S": i["quadkey"]},
-                                }
-                            }
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.batch_write_item
+        max_items = 25
+        for c in _chunks(resp["Items"], max_items):
+            items = [
+                {
+                    "DeleteRequest": {
+                        "Key": {
+                            "mosaicId": {"S": self.mosaic_name},
+                            "quadkey": {"S": item["quadkey"]},
                         }
-                    ]
+                    }
                 }
-            )
+                for item in c
+            ]
+            self.client.batch_write_item(RequestItems={self.table_name: items})
--- a/cogeo_mosaic/backends/dynamodb.py
+++ b/cogeo_mosaic/backends/dynamodb.py
@@ -183,8 +183,8 @@ class DynamoDBBackend(BaseBackend):
             {"AttributeName": "quadkey", "AttributeType": "S"},
         ]
         key_schema = [
-            {"AttributeName": "mosaicId", "KeyType": "RANGE"},
-            {"AttributeName": "quadkey", "KeyType": "HASH"},
+            {"AttributeName": "mosaicId", "KeyType": "HASH"},
+            {"AttributeName": "quadkey", "KeyType": "RANGE"},
         ]
 
         # Note: errors if table already exists
@@ -213,6 +213,7 @@ class DynamoDBBackend(BaseBackend):
 
         # NOTE: quadkey is a string type
         meta["quadkey"] = "-1"
+        meta["mosaicId"] = self.mosaic_name
         items.append(meta)
 
         for quadkey, assets in self.mosaic_def.tiles.items():
--- a/cogeo_mosaic/backends/dynamodb.py
+++ b/cogeo_mosaic/backends/dynamodb.py
@@ -3,6 +3,7 @@
 import itertools
 import json
 import os
+import re
 import sys
 import warnings
 from decimal import Decimal
@@ -19,30 +20,45 @@ from cachetools.keys import hashkey
 from cogeo_mosaic.backends.base import BaseBackend
 from cogeo_mosaic.backends.utils import find_quadkeys
 from cogeo_mosaic.cache import lru_cache
-from cogeo_mosaic.errors import _HTTP_EXCEPTIONS, MosaicError, MosaicExists
+from cogeo_mosaic.errors import _HTTP_EXCEPTIONS, MosaicError
 from cogeo_mosaic.mosaic import MosaicJSON
 from cogeo_mosaic.utils import bbox_union
 
 
-@attr.s
+@attr.s()
 class DynamoDBBackend(BaseBackend):
     """DynamoDB Backend Adapter."""
 
     client: Any = attr.ib(default=None)
     region: str = attr.ib(default=os.getenv("AWS_REGION", "us-east-1"))
     table_name: str = attr.ib(init=False)
+    mosaic_name: str = attr.ib(init=False)
     table: Any = attr.ib(init=False)
 
     _backend_name = "AWS DynamoDB"
 
     def __attrs_post_init__(self):
-        """Post Init: parse path, create client and connect to Table."""
+        """Post Init: parse path, create client and connect to Table.
+
+        A path looks like
+
+        dynamodb://{region}/{table_name}:{mosaic_name}
+        dynamodb:///{table_name}:{mosaic_name}
+
+        """
+        if not re.match(
+            r"dynamodb://([a-z]{2}\-[a-z]+\-[0-9])?\/[a-zA-Z0-9\_\-\.]+\:[a-zA-Z0-9\_\-\.]+$",
+            self.path,
+        ):
+            raise ValueError(f"Invalid DynamoDB path: {self.path}")
+
         parsed = urlparse(self.path)
-        self.table_name = parsed.path.strip("/")
+        self.table_name, self.mosaic_name = parsed.path.lstrip("/").split(":")
         self.region = parsed.netloc or self.region
         self.client = self.client or boto3.resource("dynamodb", region_name=self.region)
         self.table = self.client.Table(self.table_name)
         super().__attrs_post_init__()
+        attr.validate(self)
 
     def assets_for_tile(self, x: int, y: int, z: int) -> List[str]:
         """Retrieve assets for tile."""
@@ -53,46 +69,47 @@ class DynamoDBBackend(BaseBackend):
         tile = mercantile.tile(lng, lat, self.quadkey_zoom)
         return self.get_assets(tile.x, tile.y, tile.z)
 
-    def info(self, fetch_quadkeys: bool = False):
+    def info(self, quadkeys: bool = False):
         """Mosaic info."""
-        if not fetch_quadkeys:
-            warnings.warn(
-                "Returning empty quadkey list, performing full scan operation might be slow and expensive on large database."
-                "You can retrieve the list of quadkey by setting `fetch_quadkeys=True`"
-            )
-
         return {
             "bounds": self.mosaic_def.bounds,
             "center": self.mosaic_def.center,
             "maxzoom": self.mosaic_def.maxzoom,
             "minzoom": self.mosaic_def.minzoom,
             "name": self.mosaic_def.name if self.mosaic_def.name else "mosaic",
-            "quadkeys": [] if not fetch_quadkeys else self._quadkeys,
+            "quadkeys": [] if not quadkeys else self._quadkeys,
         }
 
     @property
     def _quadkeys(self) -> List[str]:
         """Return the list of quadkey tiles."""
-        warnings.warn(
-            "Performing full scan operation might be slow and expensive on large database."
+        resp = self.table.query(
+            KeyConditionExpression="#mosaicId = :mosaicId",
+            # allows you to use dyanmodb reserved keywords as field names
+            ExpressionAttributeNames={"#mosaicId": "mosaicId", "#quadKey": "quadKey"},
+            ExpressionAttributeValues={":mosaicId": {"S": self.mosaic_name}},
+            ProjectionExpression="#quadKey",
         )
-        resp = self.table.scan(ProjectionExpression="quadkey")  # TODO: Add pagination
         return [qk["quadkey"] for qk in resp["Items"] if qk["quadkey"] != "-1"]
 
     def write(self, overwrite: bool = False, **kwargs: Any):
         """Write mosaicjson document to AWS DynamoDB."""
-        self._create_table(overwrite=overwrite, **kwargs)
+        if self._table_exist() and overwrite:
+            self._create_table(overwrite=overwrite, **kwargs)
         items = self._create_items()
         self._write_items(items)
 
     def _update_quadkey(self, quadkey: str, dataset: List[str]):
-        """Update quadkey list."""
-        self.table.put_item(Item={"quadkey": quadkey, "assets": dataset})
+        """Update single quadkey in DynamoDB."""
+        self.table.put_item(
+            Item={"mosaicId": self.mosaic_name, "quadkey": quadkey, "assets": dataset}
+        )
 
     def _update_metadata(self):
         """Update bounds and center."""
         meta = json.loads(json.dumps(self.metadata), parse_float=Decimal)
         meta["quadkey"] = "-1"
+        meta["mosaicId"] = self.mosaic_name
         self.table.put_item(Item=meta)
 
     def update(
@@ -136,21 +153,19 @@ class DynamoDBBackend(BaseBackend):
 
         self._update_metadata()
 
-        return
-
     def _create_table(
         self, overwrite: bool = False, billing_mode: str = "PAY_PER_REQUEST"
     ):
-        if self._table_exist():
-            if not overwrite:
-                raise MosaicExists("Table already exist, use `overwrite=True`.")
-            self.table.delete()
-            self.table.wait_until_not_exists()
-
         # Define schema for primary key
         # Non-keys don't need a schema
-        attr_defs = [{"AttributeName": "quadkey", "AttributeType": "S"}]
-        key_schema = [{"AttributeName": "quadkey", "KeyType": "HASH"}]
+        attr_defs = [
+            {"AttributeName": "mosaicId", "AttributeType": "S"},
+            {"AttributeName": "quadkey", "AttributeType": "S"},
+        ]
+        key_schema = [
+            {"AttributeName": "mosaicId", "KeyType": "RANGE"},
+            {"AttributeName": "quadkey", "KeyType": "HASH"},
+        ]
 
         # Note: errors if table already exists
         try:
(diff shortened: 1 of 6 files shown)


cogeo_mosaic/backends/dynamodb.py
```diff
@@ -53,46 +69,47 @@ def assets_for_point(self, lng: float, lat: float) -> List[str]:
         tile = mercantile.tile(lng, lat, self.quadkey_zoom)
         return self.get_assets(tile.x, tile.y, tile.z)
 
-    def info(self, fetch_quadkeys: bool = False):
+    def info(self, quadkeys: bool = False):
```
vincentsarago, line 72: will be the same for all the backend once we do #126

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -31,18 +32,33 @@ class DynamoDBBackend(BaseBackend):
...
+        """
+        if not re.match(
+            r"dynamodb://([a-z]{2}\-[a-z]+\-[0-9])?\/[a-zA-Z0-9\_\-\.]+\:[a-zA-Z0-9\_\-\.]+$",
+            self.path,
+        ):
+            raise ValueError(f"Invalid DynamoDB path: {self.path}")
+
         parsed = urlparse(self.path)
-        self.table_name = parsed.path.strip("/")
+        self.table_name, self.mosaic_name = parsed.path.lstrip("/").split(":")
```
vincentsarago, line 56: maybe two lines here will be clearer!

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -31,18 +32,33 @@ class DynamoDBBackend(BaseBackend):
...
         self.region = parsed.netloc or self.region
         self.client = self.client or boto3.resource("dynamodb", region_name=self.region)
         self.table = self.client.Table(self.table_name)
         super().__attrs_post_init__()
+        attr.validate(self)
```
vincentsarago, line 61: to be removed

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -53,46 +69,47 @@ def assets_for_point(self, lng: float, lat: float) -> List[str]:
...
+            ProjectionExpression="#quadKey",
         )
-        resp = self.table.scan(ProjectionExpression="quadkey")  # TODO: Add pagination
         return [qk["quadkey"] for qk in resp["Items"] if qk["quadkey"] != "-1"]
 
     def write(self, overwrite: bool = False, **kwargs: Any):
         """Write mosaicjson document to AWS DynamoDB."""
-        self._create_table(overwrite=overwrite, **kwargs)
+        if self._table_exist() and overwrite:
+            self._create_table(overwrite=overwrite, **kwargs)
```
vincentsarago, line 98: I think the overwrite should do something else here. e.g check if the mosaic already exist and if yes, clean the items and write the new one.

`write` should only `create_table` if it doesn't exits

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -136,21 +153,19 @@ def update(
 
         self._update_metadata()
 
-        return
-
     def _create_table(
         self, overwrite: bool = False, billing_mode: str = "PAY_PER_REQUEST"
```
vincentsarago, line 157: get ride of overwrite here!

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -244,3 +271,11 @@ def _table_exist(self) -> bool:
             return True
         except self.table.meta.client.exceptions.ResourceNotFoundException:
             return False
+
+    def _mosaic_exist(self) -> bool:
+        """Check if the mosaic already Exist in the Table."""
+        pass
```
vincentsarago, line 277: how to?

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -244,3 +271,11 @@ def _table_exist(self) -> bool:
...
+
+    def clean(self):
+        """clean MosaicID from dynamoDB Table."""
+        pass
```
vincentsarago, line 281: how to ?

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -244,3 +271,11 @@ def _table_exist(self) -> bool:
             return True
         except self.table.meta.client.exceptions.ResourceNotFoundException:
             return False
+
+    def _mosaic_exist(self) -> bool:
```
vincentsarago, line 275: exist or exists ?

leothomas: exists IMO

cogeo_mosaic/backends/dynamodb.py
leothomas: There is not easy way to check if a partition key/hash key is in a table. One if all mosaic's are guarenteed to have a metadata object stored under `quadKey = "-1"` I would recommend doing a `get_item({mosaicId = "...", "quadKey" = "-1"}` otherwise `putItem` and `updateItem` operations can be restricted to execute only if they meet a certain [condition](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html).

vincentsarago: > I would recommend doing a get_item({mosaicId = "...", "quadKey" = "-1"}
👍

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -136,21 +172,20 @@ def update(
...
-        attr_defs = [{"AttributeName": "quadkey", "AttributeType": "S"}]
-        key_schema = [{"AttributeName": "quadkey", "KeyType": "HASH"}]
+        attr_defs = [
+            {"AttributeName": "mosaicId", "AttributeType": "S"},
+            {"AttributeName": "quadkey", "AttributeType": "S"},
+        ]
+        key_schema = [
+            {"AttributeName": "mosaicId", "KeyType": "RANGE"},
+            {"AttributeName": "quadkey", "KeyType": "HASH"},
+        ]
```
vincentsarago, line 188: @leothomas I'm getting 
> ClientError: An error occurred (ValidationException) when calling the CreateTable operation: Invalid KeySchema: The first KeySchemaElement is not a HASH key type

any idea?

cogeo_mosaic/backends/dynamodb.py
leothomas: Unfortunately it's not possible to delete all items belonging to a single partition/hash key. I would recommend querying `_quadKeys` and doing a [batchWriteItem](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.batch_write_item) operation using the `DeleteRequest` Parameter

vincentsarago: check the latest commit ;-)

cogeo_mosaic/backends/dynamodb.py
leothomas: Could it be just switching the order of the two elements in the list? (ie: 
```
key_schema = [
            {"AttributeName": "quadkey", "KeyType": "HASH"},
            {"AttributeName": "mosaicId", "KeyType": "RANGE"},
        ]
```
)

vincentsarago: yep tried and it worked at least for creating the table. Now I have another error 🎉

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -180,7 +216,7 @@ def _create_items(self) -> List[Dict]:
         items.append(meta)
 
         for quadkey, assets in self.mosaic_def.tiles.items():
-            item = {"quadkey": quadkey, "assets": assets}
+            item = {"mosaicId": self.mosaic_name, "quadkey": quadkey, "assets": assets}
```
vincentsarago, line 219: @leothomas I'm getting
> ClientError: An error occurred (ValidationException) when calling the BatchWriteItem operation: The provided key element does not match the schema

when trying to write items

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -183,8 +183,8 @@ def _create_table(self, billing_mode: str = "PAY_PER_REQUEST", **kwargs: Any):
             {"AttributeName": "quadkey", "AttributeType": "S"},
         ]
         key_schema = [
-            {"AttributeName": "mosaicId", "KeyType": "RANGE"},
-            {"AttributeName": "quadkey", "KeyType": "HASH"},
+            {"AttributeName": "mosaicId", "KeyType": "HASH"},
+            {"AttributeName": "quadkey", "KeyType": "RANGE"},
```
kylebarron, line 187: fyi quadkey is currently a hash key I think

vincentsarago: @kylebarron do you think this change is a problem ?

kylebarron: No actually this change makes sense, the hash is the partitioning and the range is a sorted index within the partition. Thus it makes sense for Mosaic id to be hash and the other range. I don't think both can be hash? https://stackoverflow.com/questions/27329461/what-is-hash-and-range-primary-key

cogeo_mosaic/backends/dynamodb.py
kylebarron: Note that in the change, Vincent switched the index types, not the order of the keys.

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -231,16 +268,55 @@ def get_assets(self, x: int, y: int, z: int) -> List[str]:
...
+        """clean MosaicID from dynamoDB Table."""
+        logger.debug(f"Deleting items for mosaic {self.mosaic_name}...")
+
+        # get items
+        resp = self.table.query(
+            KeyConditionExpression="#mosaicId = :mosaicId",
+            ExpressionAttributeNames={"#mosaicId": "mosaicId", "#quadKey": "quadKey"},
+            ExpressionAttributeValues={":mosaicId": {"S": self.mosaic_name}},
+            ProjectionExpression="#quadKey",
+        )
```
vincentsarago, line 305: this isn't working

kylebarron: IMO this isn't great since it requires a full table scan any time you want to delete records.

cogeo_mosaic/backends/dynamodb.py
leothomas: > I don't think both can be hash
Confirming that both cannot be hash key. DynamoDB records are identified by a unique primary key, which can be composed of just a hash key (which was originally the case, with `quadKey` as the hash key) or a hash + range key (which is now the case, with `mosaic_id` as hash key and `quadKey` as range key)

cogeo_mosaic/backends/dynamodb.py
leothomas: This doesn't quite do a full table scan, since it queries for a partition key (`mosaicId`) meaning that it only searches within a single partition. I agree that deleting records in a loop isn't the most efficient, @vincentsarago  mentioned that the operation would be not be performed often enough to affect costs/performance

leothomas: That being said @vincentsarago, why isn't it working?

vincentsarago: I'll try to reproduce and post the error!

kylebarron: > it queries for a partition key (`mosaicId`) meaning that it only searches within a single partition

Thanks for clearing that up, it makes a lot of sense! I was worried that a very large table with many `mosaicid`s would be excessively slow, but in effect it's pretty similar to the operation that happens currently when you delete a table.

leothomas: 💯

vincentsarago:
```
ClientError                               Traceback (most recent call last)
<ipython-input-7-dc3e028485c1> in <module>
----> 1 m.clean()

~/Workspace/DevSeed/cogeo-mosaic/cogeo_mosaic/backends/dynamodb.py in clean(self)
    302             ExpressionAttributeNames={"#mosaicId": "mosaicId", "#quadKey": "quadKey"},
    303             ExpressionAttributeValues={":mosaicId": {"S": self.mosaic_name}},
--> 304             ProjectionExpression="#quadKey",
    305         )
    306 

~/Workspace/venv/py37/lib/python3.7/site-packages/boto3/resources/factory.py in do_action(self, *args, **kwargs)
    518             # instance via ``self``.
    519             def do_action(self, *args, **kwargs):
--> 520                 response = action(self, *args, **kwargs)
    521 
    522                 if hasattr(self, 'load'):

~/Workspace/venv/py37/lib/python3.7/site-packages/boto3/resources/action.py in __call__(self, parent, *args, **kwargs)
     81                     operation_name, params)
     82 
---> 83         response = getattr(parent.meta.client, operation_name)(**params)
     84 
     85         logger.debug('Response: %r', response)

~/Workspace/venv/py37/lib/python3.7/site-packages/botocore/client.py in _api_call(self, *args, **kwargs)
    314                     "%s() only accepts keyword arguments." % py_operation_name)
    315             # The "self" in this scope is referring to the BaseClient.
--> 316             return self._make_api_call(operation_name, kwargs)
    317 
    318         _api_call.__name__ = str(py_operation_name)

~/Workspace/venv/py37/lib/python3.7/site-packages/botocore/client.py in _make_api_call(self, operation_name, api_params)
    624             error_code = parsed_response.get("Error", {}).get("Code")
    625             error_class = self.exceptions.from_code(error_code)
--> 626             raise error_class(parsed_response, operation_name)
    627         else:
    628             return parsed_response

ClientError: An error occurred (ValidationException) when calling the Query operation: One or more parameter values were invalid: Condition parameter type does not match schema type
```

vincentsarago: 🤦 `quadKey` -> `quadkey`

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -231,16 +267,53 @@ def get_assets(self, x: int, y: int, z: int) -> List[str]:
...
+                    "DeleteRequest": {
+                        "Key": {
+                            "mosaicId": {"S": self.mosaic_name},
+                            "quadkey": {"S": item["quadkey"]},
+                        }
+                    }
+                }
+                for item in c
+            ]
+            self.client.batch_write_item(RequestItems={self.table_name: items})
```
vincentsarago, line 319: This is still not working with exeption:
```
~/Workspace/DevSeed/cogeo-mosaic/cogeo_mosaic/backends/dynamodb.py in clean(self)
    315                 for item in c
    316             ]
--> 317             self.client.batch_write_item(RequestItems={self.table_name: items})

~/Workspace/venv/py37/lib/python3.7/site-packages/boto3/resources/factory.py in do_action(self, *args, **kwargs)
    518             # instance via ``self``.
    519             def do_action(self, *args, **kwargs):
--> 520                 response = action(self, *args, **kwargs)
    521 
    522                 if hasattr(self, 'load'):

~/Workspace/venv/py37/lib/python3.7/site-packages/boto3/resources/action.py in __call__(self, parent, *args, **kwargs)
     81                     operation_name, params)
     82 
---> 83         response = getattr(parent.meta.client, operation_name)(**params)
     84 
     85         logger.debug('Response: %r', response)

~/Workspace/venv/py37/lib/python3.7/site-packages/botocore/client.py in _api_call(self, *args, **kwargs)
    314                     "%s() only accepts keyword arguments." % py_operation_name)
    315             # The "self" in this scope is referring to the BaseClient.
--> 316             return self._make_api_call(operation_name, kwargs)
    317 
    318         _api_call.__name__ = str(py_operation_name)

~/Workspace/venv/py37/lib/python3.7/site-packages/botocore/client.py in _make_api_call(self, operation_name, api_params)
    624             error_code = parsed_response.get("Error", {}).get("Code")
    625             error_class = self.exceptions.from_code(error_code)
--> 626             raise error_class(parsed_response, operation_name)
    627         else:
    628             return parsed_response

ClientError: An error occurred (ValidationException) when calling the BatchWriteItem operation: The provided key element does not match the schema
```

cc @leothomas

vincentsarago: here is an example of `items`
```
[{'DeleteRequest': {'Key': {'mosaicId': {'S': 'vincents.test'},
    'quadkey': {'S': '-1'}}}},
 {'DeleteRequest': {'Key': {'mosaicId': {'S': 'vincents.test'},
    'quadkey': {'S': '0302300'}}}},
 {'DeleteRequest': {'Key': {'mosaicId': {'S': 'vincents.test'},
    'quadkey': {'S': '0302301'}}}},
 {'DeleteRequest': {'Key': {'mosaicId': {'S': 'vincents.test'},
    'quadkey': {'S': '0302302'}}}},
 {'DeleteRequest': {'Key': {'mosaicId': {'S': 'vincents.test'},
    'quadkey': {'S': '0302303'}}}},
 {'DeleteRequest': {'Key': {'mosaicId': {'S': 'vincents.test'},
    'quadkey': {'S': '0302310'}}}},
 {'DeleteRequest': {'Key': {'mosaicId': {'S': 'vincents.test'},
    'quadkey': {'S': '0302312'}}}},
 {'DeleteRequest': {'Key': {'mosaicId': {'S': 'vincents.test'},
    'quadkey': {'S': '0302320'}}}},
 {'DeleteRequest': {'Key': {'mosaicId': {'S': 'vincents.test'},
    'quadkey': {'S': '0302321'}}}},
 {'DeleteRequest': {'Key': {'mosaicId': {'S': 'vincents.test'},
    'quadkey': {'S': '0302330'}}}}]
```

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -93,11 +95,8 @@ def info(self, quadkeys: bool = False):
     def _quadkeys(self) -> List[str]:
         """Return the list of quadkey tiles."""
         resp = self.table.query(
-            KeyConditionExpression="#mosaicId = :mosaicId",
-            # allows you to use dyanmodb reserved keywords as field names
-            ExpressionAttributeNames={"#mosaicId": "mosaicId", "#quadKey": "quadKey"},
-            ExpressionAttributeValues={":mosaicId": {"S": self.mosaic_name}},
-            ProjectionExpression="#quadKey",
+            KeyConditionExpression=Key("mosaicId").eq(self.mosaic_name),
```
kylebarron, line 98: ❤️

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -296,27 +295,25 @@ def clean(self):
...
 
-        # get items
+        # get quadkeys
         resp = self.table.query(
-            KeyConditionExpression="#mosaicId = :mosaicId",
-            ExpressionAttributeNames={"#mosaicId": "mosaicId", "#quadKey": "quadKey"},
-            ExpressionAttributeValues={":mosaicId": {"S": self.mosaic_name}},
-            ProjectionExpression="#quadKey",
+            KeyConditionExpression=Key("mosaicId").eq(self.mosaic_name),
+            ProjectionExpression="quadkey",
```
kylebarron, line 301: Can you reuse `self._quadkeys` here?

vincentsarago: no because in _quadkeys we remove `-1`, we could append it 🤷‍♂️

vincentsarago: just feel that we will save 2 lines to add one back
```python
qks = self._quadkeys + ["-1"]
```

kylebarron: I think that would be simpler than copying the dynamodb query twice. It also might be a good idea to have a dummy `_meta_quadkey` function:

```py
@property
def _meta_quadkey(self):
	return "-1"
```

then you could use
```py
quadkeys = [*self._quadkeys + self._meta_quadkey]
```

There are probably some other places where `_meta_quadkey` would help for clarity.

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -296,21 +300,18 @@ def clean(self):
...
         max_items = 25
-        for c in _chunks(resp["Items"], max_items):
+        for c in _chunks(dbitems, max_items):
             items = [
                 {
                     "DeleteRequest": {
                         "Key": {
                             "mosaicId": {"S": self.mosaic_name},
-                            "quadkey": {"S": item["quadkey"]},
+                            "quadkey": {"S": c},
```
vincentsarago, line 314: 😂

cogeo_mosaic/backends/dynamodb.py
vincentsarago: using `table.batch_writer` works 🎉

tests/test_backends.py
```diff
@@ -389,31 +389,31 @@ def test_dynamoDB_backend(client):
...
-        # Warns in backend._quadkeys
-        with pytest.warns(UserWarning):
-            info = mosaic.info(fetch_quadkeys=True)
-            assert info["quadkeys"]
+        info = mosaic.info(quadkeys=True)
+        assert info["quadkeys"]
 
     with MosaicBackend(
-        "dynamodb:///thiswaskylebarronidea", mosaic_def=mosaic_content
+        "dynamodb:///thiswaskylebarronidea:mosaic2", mosaic_def=mosaic_content
```
kylebarron, line 407: 😄

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -31,14 +34,38 @@ class DynamoDBBackend(BaseBackend):
...
+        A path looks like
+
+        dynamodb://{region}/{table_name}:{mosaic_name}
+        dynamodb:///{table_name}:{mosaic_name}
+
+        """
+        logger.debug(f"Using DynamoDB backend: {self.path}")
+
+        if not re.match(
+            r"dynamodb://([a-z]{2}\-[a-z]+\-[0-9])?\/[a-zA-Z0-9\_\-\.]+\:[a-zA-Z0-9\_\-\.]+$",
```
kylebarron, line 55: So a non-empty `mosaic_name` is required right?

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -53,46 +80,56 @@ def assets_for_point(self, lng: float, lat: float) -> List[str]:
...
-            "Performing full scan operation might be slow and expensive on large database."
+        resp = self.table.query(
+            KeyConditionExpression=Key("mosaicId").eq(self.mosaic_name),
+            ProjectionExpression="quadkey",
         )
-        resp = self.table.scan(ProjectionExpression="quadkey")  # TODO: Add pagination
-        return [qk["quadkey"] for qk in resp["Items"] if qk["quadkey"] != "-1"]
+        return [
+            qk["quadkey"]
+            for qk in resp["Items"]
```
kylebarron, line 103: I think you should choose a variable name other than `qk` here. I feel like `qk` should only represent a quadkey itself, while this is more like an `item` which contains a quadkey

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -53,46 +80,56 @@ def assets_for_point(self, lng: float, lat: float) -> List[str]:
...
+            if qk["quadkey"] != self._metadata_quadkey
+        ]
 
     def write(self, overwrite: bool = False, **kwargs: Any):
         """Write mosaicjson document to AWS DynamoDB."""
```
kylebarron, line 108: Add note that `kwargs` are passed to `self._create_table`?

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -231,16 +270,37 @@ def get_assets(self, x: int, y: int, z: int) -> List[str]:
...
+
+    def _mosaic_exists(self) -> bool:
+        """Check if the mosaic already exists in the Table."""
+        item = self.table.get_item(
+            Key={"mosaicId": self.mosaic_name, "quadkey": self._metadata_quadkey}
+        ).get("Item", {})
+
+        return True if item else False
+
+    def clean(self):
```
kylebarron, line 297: I think you should rename to something like `delete`. `clean` doesn't necessarily convey that it's a destructive action.

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -53,46 +80,56 @@ def assets_for_point(self, lng: float, lat: float) -> List[str]:
...
 
     def write(self, overwrite: bool = False, **kwargs: Any):
         """Write mosaicjson document to AWS DynamoDB."""
-        self._create_table(overwrite=overwrite, **kwargs)
+        if not self._table_exists():
+            self._create_table(**kwargs)
+
+        if self._mosaic_exists():
+            if not overwrite:
+                raise MosaicExists(
```
kylebarron, line 114: Should errors usually end in `Error`? I.e. seems like it would be better suited to be named `MosaicExistsError`

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -53,46 +80,56 @@ def assets_for_point(self, lng: float, lat: float) -> List[str]:
...
-        """Update quadkey list."""
-        self.table.put_item(Item={"quadkey": quadkey, "assets": dataset})
+        """Update single quadkey in DynamoDB."""
+        self.table.put_item(
+            Item={"mosaicId": self.mosaic_name, "quadkey": quadkey, "assets": dataset}
+        )
 
     def _update_metadata(self):
         """Update bounds and center."""
         meta = json.loads(json.dumps(self.metadata), parse_float=Decimal)
```
kylebarron, line 130: Probably good to add a comment like "Note: this is required because DynamoDB requires all numbers to be in Decimal type"

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -136,21 +175,20 @@ def update(
...
-    def _create_table(
-        self, overwrite: bool = False, billing_mode: str = "PAY_PER_REQUEST"
-    ):
-        if self._table_exist():
-            if not overwrite:
-                raise MosaicExists("Table already exist, use `overwrite=True`.")
-            self.table.delete()
-            self.table.wait_until_not_exists()
+    def _create_table(self, billing_mode: str = "PAY_PER_REQUEST", **kwargs: Any):
+        """Create DynamoDB Table."""
```
kylebarron, line 179: Probably good to document that `**kwargs` get passed on to `self.client.create_table`

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -175,12 +214,12 @@ def _create_items(self) -> List[Dict]:
         # https://blog.ruanbekker.com/blog/2019/02/05/convert-float-to-decimal-data-types-for-boto3-dynamodb-using-python/
```
kylebarron, line 214: ^^ this fn could use a good docstring (sorry)

```
Create DynamoDB items from Mosaic definition
```

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -231,16 +270,37 @@ def get_assets(self, x: int, y: int, z: int) -> List[str]:
...
         except self.table.meta.client.exceptions.ResourceNotFoundException:
             return False
+
+    def _mosaic_exists(self) -> bool:
+        """Check if the mosaic already exists in the Table."""
+        item = self.table.get_item(
+            Key={"mosaicId": self.mosaic_name, "quadkey": self._metadata_quadkey}
+        ).get("Item", {})
+
+        return True if item else False
```
kylebarron, line 295: Could just do `return bool(item)`

cogeo_mosaic/backends/dynamodb.py
vincentsarago: yes

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -294,9 +317,9 @@ def _mosaic_exists(self) -> bool:
 
         return True if item else False
 
-    def clean(self):
-        """clean MosaicID from dynamoDB Table."""
-        logger.debug(f"Deleting items for mosaic {self.mosaic_name}...")
+    def delete(self):
+        """Delete every items for a specific mosaic in the dynamoDB Table."""
```
kylebarron, line 320: `Delete all items`

kylebarron (approved):

cogeo_mosaic/backends/dynamodb.py
vincentsarago, line 55: missing `^`

PR #128, opened 2020-10-20
add CLI to create GeoJSON from mosaic

vincentsarago:


cogeo_mosaic/scripts/cli.py
```diff
@@ -420,3 +421,48 @@ def info(input, to_json):
...
         )
+
+
+@cogeo_cli.command(short_help="Create GeoJSON from a MosaicJSON document")
+@click.argument("input", type=click.Path())
+@click.option(
+    "--collect",
+    is_flag=True,
+    default=False,
+    help="Output as a GeoJSON feature collections.",
```
kylebarron, line 432: Might be good to say that when `False`, output is newline-delimited GeoJSON

PR #133, opened 2020-10-22
Rename http.py -> web.py

kylebarron:
It's probably good to rename `http.py` to another file name because a Python stdlib module is named `http`. So for example I can't even start a Jupyter kernel/session in that directory, because the local `http.py` hides the stdlib import.

Completely open to suggestions of how to rename `http.py` 🤷‍♂️


vincentsarago:
I think web is fine, thanks @kylebarron

PR #145, opened 2020-12-14
removes rio_tiler.mercator

vincentsarago:


cogeo_mosaic/utils.py
```diff
@@ -42,11 +40,8 @@ def _filter_futures(tasks):
 
 def get_dataset_info(src_path: str) -> Dict:
     """Get rasterio dataset meta."""
-    with rasterio.open(src_path) as src_dst:
-        bounds = transform_bounds(
-            src_dst.crs, "epsg:4326", *src_dst.bounds, densify_pts=21
-        )
-        min_zoom, max_zoom = get_zooms(src_dst, ensure_global_max_zoom=True)
+    with COGReader(src_path) as cog:
+        bounds = cog.bounds
```
sdfgeoff, line 44: Note that this removal breaks changes for any geotiffs that aren't already in `epsg:4326` CRS.

1. Run `MosaicJSON.from_urls` and have a geotiff that is in a funky CRS
2. The output mosaicJSON silently does not include said geotiff in any tiles.

vincentsarago: this was fixed later https://github.com/developmentseed/cogeo-mosaic/blob/master/cogeo_mosaic/utils.py#L47

Issue #142, opened 2020-12-07
Explore other Mosaic Backends (DB)


vincentsarago:
DynamoDB backend is really interesting because it enables large-scale mosaic and with the latest changes we can even store a lot of mosaics in a single table. 

As Mosaic become more popular (at least on my side 😄) I'll need to explore other `DB` backends for other cloud providers (e,g Azure Cosmos)

vincentsarago:
I'm also thinking about `read only` backends. 

By `read only` I'm mean, we could connect cogeo-mosaic backends to spatial database (or spatial index). Those won't have to have all MosaicJSON implementation but we could just use `get_assets` to do the spatial query.


```python
with RTreeBackend("spatial_index.r") as mosaic:
    mosaic.assets_for_tile(1,2,3)
    mosaic.assets_for_point(lng, lat)
```

kylebarron:
Permitting read-only backends could allow you to make a read-only adapter to query a postgres-backed STAC API...

vincentsarago:
see: https://github.com/stac-utils/titiler-pgstac/blob/master/titiler/pgstac/mosaic.py or https://github.com/developmentseed/cogeo-mosaic/blob/master/docs/examples/Create_a_Dynamic_RtreeBackend.ipynb

PR #148, opened 2021-01-13
add SQLite Backend

vincentsarago:
~~The SQLite MosaicJSON backend follows what we did in dynamoDB.~~

## `mosaicjson_metadata` Table

By default we store the mosaic metadata in a `reserved` table `mosaicjson_metadata`. It's schema is following the mosaicJSON specification (minus the `tiles`).

### Schema

row | type
-- | --
mosaicjson | TEXT NOT NULL
name | TEXT NOT NULL
description | TEXT
version | TEXT NOT NULL
attribution | TEXT
minzoom | INTEGER NOT NULL
maxzoom | INTEGER NOT NULL
quadkey_zoom | INTEGER
bounds | JSON NOT NULL
center | JSON

## `{mosaic}` Table

When creating a mosaic, the backend will update the `mosaicjson_metadata` table with the metadata and update a new table with the `mosaic_name` provided. This table will be used to store the `tiles` info ({quadkey: [datasets]})

### Schema

row | type
-- | --
quadkey | TEXT NOT NULL,
assets | JSON NOT NULL

### path and mosaic name

The database can old multiple mosaics thus the user needs to pass it in the path name
```
{schema}:///{path to sqlite file}:{mosaic_name}
sqlite:///mydb.db:mosaic
```

## To Do
- [x] add tests
- [x] add docs

ref #142


cogeo_mosaic/backends/sqlite.py
```diff
@@ -0,0 +1,237 @@
...
+            quadkey_zoom=self.quadkey_zoom,
+            quiet=quiet,
+            **kwargs,
+        )
+
+        for quadkey, new_assets in new_mosaic.tiles.items():
+            tile = mercantile.quadkey_to_tile(quadkey)
+            assets = self.assets_for_tile(*tile)
+            assets = [*new_assets, *assets] if add_first else [*assets, *new_assets]
+            self._update_quadkey(quadkey, assets)
```
vincentsarago, line 170: Here we could maybe use `executemany` and also maybe to SQL operation directly 🤷‍♂️

kylebarron (commented):
I think the main reason we set `-1` to be the metadata quadkey for DynamoDB is that DynamoDB is a no-sql database, so the values are any json document. In a relational database like SQLite, I think it would be more standard to expand each key. For the `tiles` mapping, it would still only have two columns: `quadkey` and `tiles`, where `quadkey` could be an integer and `tiles` a JSONb. But I think standard schema for metadata would have it be a separate table, so that more fields could be their own columns.

cogeo_mosaic/backends/sqlite.py
```diff
@@ -0,0 +1,237 @@
...
+                raise MosaicExistsError(
+                    f"'{self.mosaic_name}' Table already exists in {self.db_name}, use `overwrite=True`."
+                )
+            self.delete()
+
+        logger.debug(f"Creating '{self.mosaic_name}' Table in {self.db_name}.")
+        with self.db:
+            self.db.execute(
+                f"CREATE TABLE {self.mosaic_name} (quadkey TEXT NOT NULL, value JSON NOT NULL);",
+            )
```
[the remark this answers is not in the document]
vincentsarago: I just took https://docs.python.org/3/library/sqlite3.html#using-the-connection-as-a-context-manager as example 🤷‍♂️ 
I don't think it will close the connection.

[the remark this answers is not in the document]
kylebarron: I didn't know that

> Connection object used as context manager only commits or rollbacks transactions, so the connection object should be closed manually

That's why you can get by without ever doing `db.commit()`. Note that I think the commit is expensive, so you can get better performance for bulk ingest if you do your writes in chunks or all at once with a commit at the end.

cogeo_mosaic/backends/sqlite.py
```diff
@@ -0,0 +1,237 @@
...
+    sqlite3 = None  # type: ignore
+    Connection = None  # type: ignore
+    OperationalError = None  # type: ignore
+
+
+@attr.s
+class SQLiteBackend(BaseBackend):
+    """SQLite Backend Adapter."""
+
+    db_name: str = attr.ib(init=False)
```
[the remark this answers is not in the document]
vincentsarago: @kylebarron  so would you say `sqlite:///:memory:mosaic` valid or would it has to be `sqlite:///:memory::mosaic` ?

cogeo_mosaic/backends/sqlite.py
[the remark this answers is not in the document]
vincentsarago: >  Note that I think the commit is expensive, so you can get better performance for bulk ingest if you do your writes in chunks or all at once with a commit at the end.

I'm not sure to get this!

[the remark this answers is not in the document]
kylebarron: I think this

```py
for item in items:
	db.insert()
	db.commit()
```

is probably much slower than

```py
for item in items:
	db.insert()

db.commit()
```

and likewise this

```py
for item in items:
	with db:
		db.insert()
```

does a commit on each iteration, and would probably be slower than

```py
with db:
	for item in items:
		db.insert()
```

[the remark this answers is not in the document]
vincentsarago: oh this is only for `update` part, not in general.

cogeo_mosaic/backends/sqlite.py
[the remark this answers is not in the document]
kylebarron: I think the latter

cogeo_mosaic/backends/sqlite.py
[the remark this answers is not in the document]
kylebarron: Well in this case most methods only do one command at a time. It's relevant mostly for bulk transactions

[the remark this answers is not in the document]
vincentsarago: I'm not sure I get this fully.

vincentsarago:
@kylebarron 
> I think the main reason we set -1 to be the metadata quadkey for DynamoDB is that DynamoDB is a no-sql database, so the values are any json document. In a relational database like SQLite, I think it would be more standard to expand each key. For the tiles mapping, it would still only have two columns: quadkey and tiles, where quadkey could be an integer and tiles a JSONb. But I think standard schema for metadata would have it be a separate table, so that more fields could be their own columns.

I agree this is quite of a hack, so following what you are saying we could create 2 tables, using the current `mosaic_name` as a prefix:
- `{mosaic_name}_metadata`
- `{mosaic_name}_tiles`

my only worry is that we have to hard code the schema of MosaicJSON, meaning that any change in the mosaicJSON will need to be reflected here to.

kylebarron:
> I agree this is quite of a hack, so following what you are saying we could create 2 tables, using the current `mosaic_name` as a prefix:

No, I was suggesting one global metadata table for _all mosaics_, plus one data table for each mosaic. E.g. geopackage has `gpkg_metadata` as the global table, plus a table for each dataset layer.

So a global metadata table like `mosaicjson_metadata` with one row of metadata for each table, plus a `{mosaic_name}` table for each mosaic.

> my only worry is that we have to hard code the schema of MosaicJSON, meaning that any change in the mosaicJSON will need to be reflected here to.

Yes... With relational databases you need to have a stable schema, or otherwise deal with database migration when your schema changes.

You could still have a global metadata table + a data table for each mosaic, while still storing all fields as jsonb 🤷‍♂️

cogeo_mosaic/backends/sqlite.py
```diff
@@ -0,0 +1,314 @@
...
+        ), f"'{self._metadata_table}' is a reserved table name."
+
+        self.db_path = path.replace(f":{self.mosaic_name}", "")
+
+        self.db = sqlite3.connect(self.db_path, detect_types=sqlite3.PARSE_DECLTYPES)
+        self.db.row_factory = sqlite3.Row
+
+        # Here we make sure the mosaicJSON.name is the same
+        if self.mosaic_def:
+            self.mosaic_def.name = self.mosaic_name
```
vincentsarago, line 67: because we use the `name` later in the SQL query we have to make sure the `name` is the same as the name provided in the uri

cogeo_mosaic/backends/sqlite.py
```diff
@@ -0,0 +1,314 @@
...
+                        :attribution,
+                        :minzoom,
+                        :maxzoom,
+                        :quadkey_zoom,
+                        :bounds,
+                        :center
+                    );
+                """,
+                self.metadata.dict(),
+            )
```
vincentsarago, line 170: This is a bit verbose, maybe there is a better way to write this

cogeo_mosaic/backends/sqlite.py
kylebarron: Might be good to add a warning here. Users might be surprised name changed?

cogeo_mosaic/backends/sqlite.py
kylebarron: 👍 to named, not positional, arguments

cogeo_mosaic/backends/base.py
```diff
@@ -101,7 +101,7 @@ def assets_for_point(self, lng: float, lat: float) -> List[str]:
 
     @cached(
         TTLCache(maxsize=cache_config.maxsize, ttl=cache_config.ttl),
-        key=lambda self, x, y, z: hashkey(self.path, x, y, z),
+        key=lambda self, x, y, z: hashkey(self.path, x, y, z, self.mosaicid),
```
vincentsarago, line 104: unrelated to the SQLite backend, but while writing the tests I realized that we were caching this even when the mosaic changed... to fix it we can use the mosaicid (built from the metadata, e.g mosaic version) to make sure we call the function when the mosaic has been updated.

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -151,7 +152,7 @@ def update(
         """Update existing MosaicJSON on backend."""
         logger.debug(f"Updating {self.mosaic_name}...")
 
-        new_mosaic = self.mosaic_def.from_features(
+        new_mosaic = MosaicJSON.from_features(
```
vincentsarago, line 155: I think using `self.mosaic_def.from_features(` was a bit misleading

cogeo_mosaic/backends/sqlite.py
```diff
@@ -0,0 +1,353 @@
...
+            not self.mosaic_name == self._metadata_table
+        ), f"'{self._metadata_table}' is a reserved table name."
+
+        self.db_path = uri_path.replace(f":{self.mosaic_name}", "")
+
+        # When mosaic_def is not passed, we have to make sure the db exists
+        if not self.mosaic_def and not os.path.exists(self.db_path):
+            raise MosaicNotFoundError(
+                f"SQLite database not found at path {self.db_path}."
+            )
```
vincentsarago, line 68: do not create a file if `mosaic_def` is not passed and file doesn't exist yet

cogeo_mosaic/backends/sqlite.py
```diff
@@ -0,0 +1,366 @@
...
+
+        self.mosaic_name = uri_path.split(":")[-1]
+        assert (
+            not self.mosaic_name == self._metadata_table
+        ), f"'{self._metadata_table}' is a reserved table name."
+
+        self.db_path = uri_path.replace(f":{self.mosaic_name}", "")
+
+        # When mosaic_def is not passed, we have to make sure the db exists
+        if not self.mosaic_def and not os.path.exists(self.db_path):
```
kylebarron, line 64: Use nicer `pathlib` API? 😄 `Path(self.db_path).exists()`?

docs/advanced/backends.md
```diff
@@ -26,20 +30,23 @@ with MosaicBackend("https://mosaic.com/amosaic.json.gz") as mosaic:
...
+
+# Read-Only
+with MosaicBackend("stac+https://my-stac.api/search", {"collections": ["satellite"]}, 10, 12) as mosaic:
+    assert isinstance(mosaic, cogeo_mosaic.backends.stac.STACBackend)
 ```
 
 ## STAC Backend
 
-The STACBackend is purely dynamic, meaning it's not used to read or write a file. This backend will POST to the input url looking for STAC items which will then be used to create the mosaicJSON.
+The STACBackend is a **read-only** backend, meaning it's not used to read or write a file. This backend will POST to the input url looking for STAC items which will then be used to create the mosaicJSON in memory.
```
kylebarron, line 49:
```suggestion
The STACBackend is a **read-only** backend, meaning it can't be used to write a file. This backend will POST to the input url looking for STAC items which will then be used to create the mosaicJSON in memory.
```

kylebarron (approved):

PR #149, opened 2021-01-14
Do URI parsing in MosaicBackend

vincentsarago:
This PR change the URI format for SQLite 

```python
# before
sqlite:///mosaic.db:mosaic_name

# now
sqlite:///mosaic.db?mosaic=mosaic_name
```

In addition the URI parsing is deferred to `cogeo_mosaic.backends.MosaicBackend`. The `SQLiteBackend` now accept `path and mosaic` as parameter

```python
# before 
SQLiteBackend("sqlite:///mosaic.db:mosaic")

# now
SQLiteBackend("mosaic.db", "mosaic")
```

## To Do
- [ ] validate with @kylebarron and @geospatial-jeff 👀


cogeo_mosaic/backends/__init__.py
```diff
@@ -16,24 +16,40 @@ def MosaicBackend(url: str, *args: Any, **kwargs: Any) -> BaseBackend:
...
 
-    if parsed.scheme == "s3":
+    # `s3:///{bucket}{key}`
+    elif parsed.scheme == "s3":
         return S3Backend(url, *args, **kwargs)
 
-    if parsed.scheme == "dynamodb":
+    # `dynamodb://{region}/{table}:{mosaic}`
+    elif parsed.scheme == "dynamodb":
         return DynamoDBBackend(url, *args, **kwargs)
```
vincentsarago, line 30: will do the same for DynamoDB in another PR

kylebarron:
I'm just kinda confused... why? you just want a named parameter?

cogeo_mosaic/backends/__init__.py
```diff
@@ -16,24 +16,40 @@ def MosaicBackend(url: str, *args: Any, **kwargs: Any) -> BaseBackend:
...
 
-    if parsed.scheme in ["https", "http"]:
+        return SQLiteBackend(db_path, qs["mosaic"], *args, **kwargs)
+
+    # https://{hostname}/{path}
+    elif parsed.scheme in ["https", "http"]:
         return HttpBackend(url, *args, **kwargs)
 
-    if parsed.scheme == "file":
+    # file://{path}
```
kylebarron, line 45:
```suggestion
    # file:///{path}
```
?

vincentsarago:
@kylebarron 

> I'm just kinda confused... why? you just want a named parameter?

I think it's still important that `MosaicBackend` accept URI like parameter. the other named parameter pithing the backends are mostly linked to the COGReader.

geospatial-jeff:
I don't think using a query parameter is any more URI than `:`, they are both supported by the rfc.  The important part to me is consistency, in my opinion `sqlite:///mosaic.db/mosaic` is best here to be consistent with the current dynamodb implementation, and then maybe using query parameters for mosaic names can be a separate issue if @vincentsarago thinks that could add value.

vincentsarago:
@geospatial-jeff 

> in my opinion sqlite:///mosaic.db/mosaic is best here to be consistent with the current dynamodb 

we use `:` in DynamoDB

https://github.com/developmentseed/cogeo-mosaic/blob/a11e36659861ef74be0ffb819c88686bef21cfa9/cogeo_mosaic/backends/dynamodb.py#L60

vincentsarago:
@kylebarron @geospatial-jeff I'll update the goal/title of the PR! I won't change the URI schema but just defer the parsing in `MosaicBackend` instead in the backends

vincentsarago:
changed my mind, I think it's ok as it is right now!

PR #150, opened 2021-01-15
add dynamic backend registration class

vincentsarago:
maybe this is totally overkill! 🤷‍♂️ 

```python
from cogeo_mosaic.backends import backends

class MyCustomBackend(BaseBackend):
    ...

backends.register("local", MyCustomBackend, lambda x: urlparse(x).path)
```

### DIFF
--- a/cogeo_mosaic/backends/__init__.py
+++ b/cogeo_mosaic/backends/__init__.py
@@ -1,6 +1,6 @@
 """cogeo_mosaic.backends."""
 
-from typing import Any
+from typing import Any, Callable, Dict, Sequence, Type, Union
 from urllib.parse import urlparse
 
 from cogeo_mosaic.backends.base import BaseBackend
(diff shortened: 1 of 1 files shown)


kylebarron:
It's not clear to me what the point of this is. What does this achieve over existing code?

vincentsarago:
@kylebarron this is to enable dynamic registration of Backend 

```python
from cogeo_mosaic.backends import backends

class MyCustomBackend(BaseBackend):
    ...

backends.register("local", MyCustomBackend, lambda x: urlparse(x).path)
```


with this we can still keep **one** top level method `cogeo_mosaic.backends.MosaicBackend` but have more backends available without needing to having them in cogeo-mosaic module


This is a bit linked to #142

vincentsarago:
alright, let's close this and maybe revisit later

PR #151, opened 2021-01-24
WIP Refactor overview code

kylebarron:
This is a WIP PR to refactor and simplify code to create COG overviews given an input mosaic. This isn't tested yet and probably has a few bugs but is enough to convey the gist.

I think it makes sense for this to go here, because the function previously lived here, and it's generally useful code. If it's not production ready, we can just expose it through an `experimental` module: `from cogeo_mosaic.experimental import create_overview_cog`.

- Overview creation always happens on disk in a temporary file. For small mosaics, this shouldn't make much of a difference in performance, and for large mosaics, overview creation would take much too much memory to happen in memory.
- Removes threading. Given my experiences in the past, I think using multiple threads would only work if you could guarantee that no two threads read the same COG at the same time. Otherwise you'd get the empty black sections of no-data that I've seen before. This is pretty hard to prove, and the `shuffle` done previously really isn't enough to guarantee anything.
- Uses `rio_cogeo.cog_translate` to create the COG after the windowed geotiff has been written once, to remove overview-writing code here
- Always creates a single output file, instead of potentially making separate files (the `split_extrema` function previously). If a user doesn't want the entire mosaic into one output COG, the user can split up the mosaic ahead of time. An example of that can go in the docs.

To do:

- [x] Restore mosaic selection method
- [x] Restore progress bar
- Future: use morecantile?

### DIFF
--- a/cogeo_mosaic/overview.py
+++ b/cogeo_mosaic/overview.py
@@ -3,7 +3,7 @@
 import random
 import sys
 from tempfile import NamedTemporaryFile
-from typing import Dict, Optional
+from typing import Dict, Optional, Union
 
 import click
 import mercantile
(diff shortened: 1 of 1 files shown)


vincentsarago:
👋 @kylebarron thanks for this PR.

As I mentioned on our private slack, I'll be happy to have an overview API/CLI back but not in cogeo-mosaic repo. We created https://github.com/developmentseed/cogeo-mosaic-plugins specifically for those kind of use case. I'd love to keep cogeo-mosaic as simple as possible and focused on reading/writing mosaicJSON. I know we disagree on this but plus try to understand my position. 

Side notes:
- this PR adds a new requirement (rio-cogeo)
- because you are using rio-cogeo you are creating 2 temporary files (that's why the code was a bit complex but it was to avoid replicating file read/write)

I do recognize that your version is way cleaner and simple that the one we had before and I'll be happy to review it when you think it's ready!

vincentsarago:
it's time to let go!

PR #157, opened 2021-01-27
[WIP] new access mode

### LINKED ISSUE #156: Change write method and input parameters for Backends.
Right now we are using a weird approach when we want to write a new mosaicJSON

```python
with MosaicBackend("apath.json.gz", mosaic_def=myMosaic) as mosaic:
     mosaic.write()
``` 

With #152, we, in theory, won't need to do stuff like 
```python
with MosaicBackend(None, mosaic_def=myMosaic) as mosaic:
     mosaic.tile(...)
```

But use the InMemoryBackend 
```python
with MosaicBackend(mosaicJSON) as mosaic:
    mosaic.tile(...) 
```

So I propose we remove the `mosaic_def` input parameters in the backends and do like 

```python
with MosaicBackend("apath.json.gz") as mosaic:
     mosaic.write(myMosaic)
```
which seems more pythonic (e.g like rasterio, file writing...) 

vincentsarago:
closes #156 

This PR refactor the way Mosaic Backends inputs works. 

```python
# Write Mosaic
### before 
with FileBackend("my_path.json.gz", mosaic_def=mymosaic) as mosaic:
    mosaic.write()

### now
with FileBackend("my_path.json.gz", mode="w") as mosaic:
    mosaic.write(mymosaic)
```

```python
# Read Mosaic
### before
with FileBackend("my_path.json.gz") as mosaic:
     ...

### now
with FileBackend("my_path.json.gz", mode="r") as mosaic: # mode is optional because we default to 'r'
     ...
```

```python
# Update Mosaic
### before
with FileBackend("my_path.json.gz") as mosaic:
     mosaic.update(anothermosaic)

### now
with FileBackend("my_path.json.gz", mode="r+") as mosaic:
         mosaic.update(anothermosaic)
```

### DIFF
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -31,13 +31,13 @@ class BaseBackend(BaseReader):
     Attributes:
         path (str): mosaic path.
         mode (str, optional): Access mode. Available modes are: r (read-only), r+ (read and write), w (write). Defaults to read-only `r`.
-        reader (rio_tiler.io.BaseReader): Dataset reader. Defaults to rio_tiler.io.COGReader.
+        reader (rio_tiler.io.BaseReader): Dataset reader. Defaults to `rio_tiler.io.COGReader`.
         reader_options (dict): Options to forward to the reader config.
         backend_options (dict): Global backend options.
         tms (morecantile.TileMatrixSet, optional): TileMatrixSet grid definition. **READ ONLY attribute**. Defaults to `WebMercatorQuad`.
-        bbox (tuple): mosaic bounds (left, bottom, right, top). **READ ONLY attribute**. Defaults to TMS's bounds.
-        minzoom (int): mosaic Min zoom level. **READ ONLY attribute**. Defaults to TMS's minzoom.
-        maxzoom (int): mosaic Max zoom level. **READ ONLY attribute**. Defaults to TMS's maxzoom.
+        bbox (tuple): mosaic bounds (left, bottom, right, top). **READ ONLY attribute**. Defaults to `(-180, -90, 180, 90)`.
+        minzoom (int): mosaic Min zoom level. **READ ONLY attribute**. Defaults to `0`.
+        maxzoom (int): mosaic Max zoom level. **READ ONLY attribute**. Defaults to `30`
         mosaic_def (MosaicJSON): mosaicJSON document. **READ ONLY attribute**.
 
     """
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -2,6 +2,7 @@
 
 import abc
 import itertools
+from copy import deepcopy
 from typing import Any, Dict, List, Optional, Sequence, Tuple, Type
 
 import attr
@@ -24,6 +25,14 @@ from cogeo_mosaic.mosaic import MosaicJSON
 from cogeo_mosaic.utils import bbox_union
 
 
+def update_zoom_and_bounds(instance, attribute, value):
+    """Update Zooms and Bounds attribute when mosaic_def is set."""
+    instance.minzoom = value.minzoom
+    instance.maxzoom = value.maxzoom
+    instance.bounds = value.bounds
+    return value
+
+
 @attr.s
 class BaseBackend(BaseReader):
     """Base Class for cogeo-mosaic backend storage.
@@ -43,14 +52,16 @@ class BaseBackend(BaseReader):
     """
 
     path: str = attr.ib()
-    mode: str = attr.ib("r")
     reader: Type[BaseReader] = attr.ib(default=COGReader)
     reader_options: Dict = attr.ib(factory=dict)
     backend_options: Dict = attr.ib(factory=dict)
+    mode: str = attr.ib(default="r")
 
     # TMS is outside the init because mosaicJSON and cogeo-mosaic only
     # works with WebMercator (mercantile) for now.
     tms: TileMatrixSet = attr.ib(init=False, default=WEB_MERCATOR_TMS)
+
+    mosaic_def: MosaicJSON = attr.ib(init=False, on_setattr=update_zoom_and_bounds)  # type: ignore
 
     bounds: Tuple[float, float, float, float] = attr.ib(
         init=False, default=(-180, -90, 180, 90)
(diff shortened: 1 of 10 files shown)


cogeo_mosaic/backends/file.py
```diff
@@ -20,12 +19,9 @@ class FileBackend(BaseBackend):
 
     _backend_name = "File"
 
-    def write(self, overwrite: bool = False, gzip: bool = None):
+    def write(self, mosaic: MosaicJSON, gzip: bool = None):
```
vincentsarago, line 22: here we have to check if the mosaic is not opened in `r` mode

cogeo_mosaic/backends/base.py
```diff
@@ -38,21 +52,34 @@ class BaseBackend(BaseReader):
...
         """Post Init: if not passed in init, try to read from self.path."""
-        self.mosaic_def = self.mosaic_def or self._read(**self.backend_options)
+        # here we can warns if the mosaic already exists and is openned in `w` mode.
+        # if self.mode == "w" and self.mosaic_exits:
+        #     raise MosaicExistsError("Mosaic already exist, use `overwrite=True`.")
 
-        self.minzoom = self.mosaic_def.minzoom
-        self.maxzoom = self.mosaic_def.maxzoom
-        self.bounds = self.mosaic_def.bounds
+        if self.mode in ["r", "r+"]:
```
kylebarron, line 77: I think you should just do `if 'r' in self.mode`. For example, isn't `rw` a valid mode?

cogeo_mosaic/backends/base.py
```diff
@@ -196,6 +223,11 @@ def update(
         **kwargs,
     ):
         """Update existing MosaicJSON on backend."""
+        if self.mode == "r":
```
kylebarron, line 226: `if 'w' not in self.mode`? It would be good to look up with the valid modes are to `open`.

kylebarron (commented):
Might be good paired with a `read-only` backend, because then you can error if the mode is anything other than `'r'`. I assume this is a breaking change.

cogeo_mosaic/backends/base.py
vincentsarago: I think I went the way it is in rasterio
`r`: read-only
`r+`: read+update
`w`: write

🤷‍♂️

cogeo_mosaic/backends/base.py
```diff
@@ -38,21 +52,34 @@ class BaseBackend(BaseReader):
...
     _backend_name: str
     _file_byte_size: Optional[int] = 0
 
-    @mosaic_def.validator
-    def _check_mosaic_def(self, attribute, value):
-        if value is not None:
-            self.mosaic_def = MosaicJSON(**dict(value))
+    @mode.validator
+    def _check_mode(self, attribute, value):
+        if value not in ["r", "r+", "w"]:
```
vincentsarago, line 68: @kylebarron here are the valid modes

vincentsarago:
@kylebarron 

> Might be good paired with a read-only backend, because then you can error if the mode is anything other than 'r'

That's the idea. 

>  I assume this is a breaking change.

absolutely.

cogeo_mosaic/backends/base.py
```diff
@@ -24,35 +25,66 @@
...
     reader: Type[BaseReader] = attr.ib(default=COGReader)
     reader_options: Dict = attr.ib(factory=dict)
     backend_options: Dict = attr.ib(factory=dict)
+    mode: str = attr.ib(default="r")
 
     # TMS is outside the init because mosaicJSON and cogeo-mosaic only
     # works with WebMercator (mercantile) for now.
     tms: TileMatrixSet = attr.ib(init=False, default=WEB_MERCATOR_TMS)
 
+    mosaic_def: MosaicJSON = attr.ib(init=False, on_setattr=update_zoom_and_bounds)  # type: ignore
```
vincentsarago, line 64: Not sure why but because of `on_setattr` I was getting typing errors

cogeo_mosaic/backends/base.py
```diff
@@ -24,35 +25,66 @@
 from cogeo_mosaic.utils import bbox_union
 
 
+def update_zoom_and_bounds(instance, attribute, value):
```
vincentsarago, line 28: magic method that get's called every time `mosaic_def` is set (updating zooms and bounds)

cogeo_mosaic/backends/base.py
```diff
@@ -196,35 +228,36 @@ def update(
...
-            self.mosaic_def.minzoom,
+            mosaic.minzoom,
         )
 
-        # We only write if path is set
-        if self.path:
-            self.write(overwrite=True)
-
-        return
+        self.write(mosaic, overwrite=True)
```
vincentsarago, line 263: here we call `write` which will write the mosaic and set `mosaic_def` to the new value

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -112,6 +114,11 @@ def write(self, overwrite: bool = False, **kwargs: Any):
             MosaicExistsError: If mosaic already exists in the Table.
 
         """
+        if self.mode == "r":
+            raise ValueError(
```
vincentsarago, line 118: maybe a custom Exception name here

cogeo_mosaic/backends/file.py
```diff
@@ -20,13 +20,21 @@ class FileBackend(BaseBackend):
 
     _backend_name = "File"
 
-    def write(self, overwrite: bool = False, gzip: bool = None):
+    def write(self, mosaic: MosaicJSON, overwrite: bool = False, gzip: bool = None):
         """Write mosaicjson document to a file."""
-        if not overwrite and os.path.exists(self.path):
+        if self.mode == "r":
+            raise ValueError(
```
vincentsarago, line 26: is ValueError good? or do we need a custom Exception?

tests/test_backends.py
```diff
@@ -403,8 +437,9 @@ def test_dynamoDB_backend(client):
     """Test DynamoDB backend."""
```
vincentsarago, line 437: DynamoDB is not well tested 😞

vincentsarago:
Maybe so much change wasn't necessary. While I like the `mode=r/r+/w+`, maybe to accommodate the `inMemory` backend we could just tell that the path has to be `:memory:`.

I'll open a new PR to explain. Basically I'll keep the `mode` but revert the fact you can pass mosaic_def in init.

vincentsarago:
I've re-worked this a bit over #160

PR #160, opened 2021-02-01
new access mode

### LINKED ISSUE #156: Change write method and input parameters for Backends.

vincentsarago:
closes #156 

The PR change the way users can write mosaicJSON using the backends. The changes also add `access mode` to the backend object to better define read/write modes. 

```python
# Write Mosaic
### before 
with FileBackend("my_path.json.gz", mosaic_def=mymosaic) as mosaic:
    mosaic.write()

### now
with FileBackend("my_path.json.gz", mode="w") as mosaic:
    mosaic.write(mymosaic)
```

For most of the backend, access mode are either:
- `r`: read-only
- `r+`: read-update
- `w`: write

All backends must have the `_available_modes` variable set to indicate which mode are supported:

https://github.com/developmentseed/cogeo-mosaic/blob/7242abc8af39eb7eebb48db195a0cc7e029e8bc3/cogeo_mosaic/backends/web.py#L27

By default the backends are opened with `r` read-only mode. 

`mosaic_def` is now outside the `init` method (**breaking**) and must be provided using the `.write()` method (if the backend is opened in `w` mode.

cc @kylebarron

### DIFF
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -2,6 +2,7 @@
 
 import abc
 import itertools
+from copy import deepcopy
 from typing import Any, Dict, List, Optional, Sequence, Tuple, Type
 
 import attr
@@ -18,18 +19,57 @@ from rio_tiler.tasks import create_tasks, filter_tasks
 
 from cogeo_mosaic.backends.utils import find_quadkeys, get_hash
 from cogeo_mosaic.cache import cache_config
-from cogeo_mosaic.errors import NoAssetFoundError
+from cogeo_mosaic.errors import NoAssetFoundError, UnsupportedOperation
 from cogeo_mosaic.models import Info, Metadata
 from cogeo_mosaic.mosaic import MosaicJSON
 from cogeo_mosaic.utils import bbox_union
 
 
+def update_zoom_and_bounds(instance, attribute, value):
+    """Update Zooms and Bounds attribute when mosaic_def is set."""
+    instance.minzoom = value.minzoom
+    instance.maxzoom = value.maxzoom
+    instance.bounds = value.bounds
+    return value
+
+
+def mode_validator(instance, attribute, value):
+    """Validate mode."""
+    if value not in instance._available_modes:
+        raise ValueError(
+            f"Invalid '{value}' mode. MUST be one of {instance._available_modes}."
+        )
+
+    if "r" in value:
+        instance._readable = True
+
+    elif "w" in value:
+        instance._writable = True
+
+    if "+" in value:
+        instance._writable = True
+
+
 @attr.s
 class BaseBackend(BaseReader):
-    """Base Class for cogeo-mosaic backend storage."""
+    """Base Class for cogeo-mosaic backend storage.
+
+    Attributes:
+        path (str): mosaic path.
+        mode (str, optional): Access mode. Available modes are: r (read-only), r+ (read and write), w (write). Defaults to read-only `r`.
+        reader (rio_tiler.io.BaseReader): Dataset reader. Defaults to `rio_tiler.io.COGReader`.
+        reader_options (dict): Options to forward to the reader config.
+        backend_options (dict): Global backend options.
+        tms (morecantile.TileMatrixSet, optional): TileMatrixSet grid definition. **READ ONLY attribute**. Defaults to `WebMercatorQuad`.
+        bbox (tuple): mosaic bounds (left, bottom, right, top). **READ ONLY attribute**. Defaults to `(-180, -90, 180, 90)`.
+        minzoom (int): mosaic Min zoom level. **READ ONLY attribute**. Defaults to `0`.
+        maxzoom (int): mosaic Max zoom level. **READ ONLY attribute**. Defaults to `30`
+        mosaic_def (MosaicJSON, optional): mosaicJSON document. **READ ONLY attribute**.
+
+    """
 
     path: str = attr.ib()
-    mosaic_def: MosaicJSON = attr.ib(default=None)
+    mode: str = attr.ib(default="r", validator=mode_validator)
     reader: Type[BaseReader] = attr.ib(default=COGReader)
     reader_options: Dict = attr.ib(factory=dict)
     backend_options: Dict = attr.ib(factory=dict)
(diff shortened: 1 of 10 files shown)


cogeo_mosaic/backends/base.py
```diff
@@ -18,18 +19,57 @@
...
+        )
+
+    if "r" in value:
+        instance._readable = True
+
+    elif "w" in value:
+        instance._writable = True
+
+    if "+" in value:
+        instance._writable = True
```
vincentsarago, line 50: Inspiration from https://github.com/python/cpython/blob/3.9/Lib/_pyio.py

cogeo_mosaic/backends/base.py
```diff
@@ -18,18 +19,57 @@
...
 from cogeo_mosaic.mosaic import MosaicJSON
 from cogeo_mosaic.utils import bbox_union
 
 
+def update_zoom_and_bounds(instance, attribute, value):
+    """Update Zooms and Bounds attribute when mosaic_def is set."""
+    instance.minzoom = value.minzoom
+    instance.maxzoom = value.maxzoom
+    instance.bounds = value.bounds
+    return value
```
vincentsarago, line 33: This is a `magic` method which will update the instance zoom/bounds each time `mosaic_def` is assigned to a new value

cogeo_mosaic/backends/base.py
```diff
@@ -38,57 +78,30 @@ class BaseBackend(BaseReader):
     # works with WebMercator (mercantile) for now.
     tms: TileMatrixSet = attr.ib(init=False, default=WEB_MERCATOR_TMS)
 
+    bounds: Tuple[float, float, float, float] = attr.ib(
+        init=False, default=(-180, -90, 180, 90)
+    )
+    minzoom: int = attr.ib(init=False, default=0)
+    maxzoom: int = attr.ib(init=False, default=30)
+
+    mosaic_def: MosaicJSON = attr.ib(init=False, on_setattr=update_zoom_and_bounds)  # type: ignore
```
vincentsarago, line 87: mosaic_def is outside the `init` method now

cogeo_mosaic/backends/base.py
```diff
@@ -196,35 +221,52 @@ def update(
...
+            mosaic.minzoom,
         )
 
-        # We only write if path is set
-        if self.path:
-            self.write(overwrite=True)
+        self.write(mosaic, overwrite=True)
+
+    ############################################################################
+    # Not Implemented methods
```
vincentsarago, line 257: I've reorganized the code a bit!

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -122,25 +167,32 @@ def write(self, overwrite: bool = False, **kwargs: Any):
...
-    def _update_metadata(self):
-        """Update bounds and center.
+        # Create Tile items
+        for quadkey, assets in self.mosaic_def.tiles.items():
+            items.append(
+                {"mosaicId": self.mosaic_name, "quadkey": quadkey, "assets": assets}
+            )
 
-        Note: `parse_float=Decimal` is required because DynamoDB requires all numbers to be in Decimal type
+        self._write_items(items)
```
vincentsarago, line 190: Here I've removed the small `_update_quadkey` and `_update_metadata`.

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -122,25 +167,32 @@ def write(self, overwrite: bool = False, **kwargs: Any):
                 )
             self.delete()
 
-        items = self._create_items()
-        self._write_items(items)
+        # Validation and Create a new instance
+        if not isinstance(mosaic, MosaicJSON):
+            mosaic = MosaicJSON(**dict(mosaic))
+        else:
+            mosaic = deepcopy(mosaic)
```
vincentsarago, line 174: because we remove `tiles` later in the method, creating a deepcopy to ensure we don't change the input mosaicJSON

vincentsarago:
```python
# Write Mosaic
### before 
with FileBackend("my_path.json.gz", mosaic_def=mymosaic) as mosaic:
    mosaic.write()

### now
with FileBackend("my_path.json.gz", mode="w") as mosaic:
    mosaic.write(mymosaic)
```

```python
# Read Mosaic
### before
with FileBackend("my_path.json.gz") as mosaic:
     ...

### now
with FileBackend("my_path.json.gz", mode="r") as mosaic: # mode is optional because we default to 'r'
     ...
```

```python
# Update Mosaic
### before
with FileBackend("my_path.json.gz") as mosaic:
     mosaic.update(anothermosaic)

### now
with FileBackend("my_path.json.gz", mode="r+") as mosaic:
         mosaic.update(anothermosaic)
```

cogeo_mosaic/backends/base.py
kylebarron: I think it might be a little clearer if this were a static method on the 
`BaseBackend` class (and maybe `mode_validator` too). That would also allow these to be subclassed if necessary for some reason.

cogeo_mosaic/backends/base.py
kylebarron: You should also add 
```
        instance._readable = True
```

This works:
```py
with open('/tmp/test.py', 'w+') as f:
    f.read()
```

cogeo_mosaic/backends/base.py
```diff
@@ -18,18 +19,57 @@
...
-    """Base Class for cogeo-mosaic backend storage."""
+    """Base Class for cogeo-mosaic backend storage.
+
+    Attributes:
+        path (str): mosaic path.
+        mode (str, optional): Access mode. Available modes are: r (read-only), r+ (read and write), w (write). Defaults to read-only `r`.
+        reader (rio_tiler.io.BaseReader): Dataset reader. Defaults to `rio_tiler.io.COGReader`.
+        reader_options (dict): Options to forward to the reader config.
+        backend_options (dict): Global backend options.
+        tms (morecantile.TileMatrixSet, optional): TileMatrixSet grid definition. **READ ONLY attribute**. Defaults to `WebMercatorQuad`.
```
kylebarron, line 63: Why does this need to be read-only?

cogeo_mosaic/backends/base.py
```diff
@@ -18,18 +19,57 @@
...
+
+    elif "w" in value:
+        instance._writable = True
+
+    if "+" in value:
+        instance._writable = True
+
+
 @attr.s
 class BaseBackend(BaseReader):
```
kylebarron, line 54: Do we still use the fact that it subclasses from `BaseReader`? There's a separate `reader` argument to the class.

cogeo_mosaic/backends/base.py
kylebarron: Why did you need to add these?

cogeo_mosaic/backends/dynamodb.py
```diff
@@ -122,25 +167,32 @@ def write(self, overwrite: bool = False, **kwargs: Any):
...
 
-    def _update_metadata(self):
-        """Update bounds and center.
+        # Create Tile items
+        for quadkey, assets in self.mosaic_def.tiles.items():
+            items.append(
+                {"mosaicId": self.mosaic_name, "quadkey": quadkey, "assets": assets}
+            )
 
-        Note: `parse_float=Decimal` is required because DynamoDB requires all numbers to be in Decimal type
```
kylebarron, line 137: Nit: I think it would be good to keep this comment around somewhere. I was surprised to need to use `Decimal`.

kylebarron (approved):

cogeo_mosaic/backends/base.py
vincentsarago: because mosaicjson doesn't support other projection than mercator for now

cogeo_mosaic/backends/base.py
vincentsarago: it has to be a subclass of rio_tilie.io.BaseReader to be compatible with other project (e.g TiTiler). The reader argument just tells the mosaic backend what reader to use for the assets

cogeo_mosaic/backends/base.py
kylebarron: Ok, why does minzoom, maxzoom need to be readonly?

cogeo_mosaic/backends/base.py
vincentsarago: let me retry but I think last time I tried those method weren't passed to subclasses, so I had to re-write the same code. having those outside the class just help for reusability

cogeo_mosaic/backends/base.py
vincentsarago: they were already there (just not the same place) and have to be in the class because of the BaseReader abstract class

cogeo_mosaic/backends/base.py
vincentsarago: because they get passed from the mosaicjson

vincentsarago:
not moving forward with this because it kinda complicate everything. 
Why? 
as pointed out by @kylebarron you cannot really do multiple `.write()` in a row, so the current flow/design is OK

PR #161, opened 2021-02-01
add inMemory Backend

### LINKED ISSUE #152: In-memory mosaic backend
If you have a mosaic in memory and you want to access backend methods like `.tile`, you currently need to know that you can pass `FileBackend(path=None, mosaic_def=...)`. It might be clearer to expose a `InMemoryBackend` that doesn't require a `path` argument.

vincentsarago:
closes #152 

I'm not sure which `flow` to choose

```python
# This PR
with InMemoryBackend(mosaicjson) as mosaic:
    ...
```

or 

```python
with InMemoryBackend(":memory:") as mosaic:
    mosaic.write(mosaicjson)
    ...
```

The second block will feel more aligned with how the other backend works, but also feels kinda weird.

Note: I didn't add the InMemoryBackend in the Top level MosaicBackend because I wasn't 💯 sure we wanted this.

### DIFF
--- a/cogeo_mosaic/backends/memory.py
+++ b/cogeo_mosaic/backends/memory.py
@@ -1,0 +1,62 @@
+"""cogeo-mosaic InMemory backend."""
+
+from typing import Dict, List, Tuple, Type
+
+import attr
+from morecantile import TileMatrixSet
+from rio_tiler.constants import WEB_MERCATOR_TMS
+from rio_tiler.io import BaseReader, COGReader
+
+from cogeo_mosaic.backends.base import (
+    BaseBackend,
+    mode_validator,
+    update_zoom_and_bounds,
+)
+from cogeo_mosaic.mosaic import MosaicJSON
+
+
+@attr.s
+class InMemoryBackend(BaseBackend):
+    """InMemory Backend Adapter
+
+    Examples:
+        >>> with InMemory(mosaicJSON) as mosaic:
+                mosaic.tile(0, 0, 0)
+
+    """
+
+    mosaic_def: MosaicJSON = attr.ib(on_setattr=update_zoom_and_bounds)  # type: ignore
+    reader: Type[BaseReader] = attr.ib(default=COGReader)
+    reader_options: Dict = attr.ib(factory=dict)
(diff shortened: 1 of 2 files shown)


kylebarron:
Do we still allow passing `mosaic_def` in the class constructor to other backends? If so, it makes sense to do it here, but maybe require it to be a keyword argument?

vincentsarago:
> Do we still allow passing mosaic_def in the class constructor to other backends? If so, it makes sense to do it here, but maybe require it to be a keyword argument?

@kylebarron no we don't, for the other backend, mosaic_def is outside the `init` and must be set using `.write(mosaic)`. IMO the InMemoryBackend is a bit different than the other backend which is why I'm fine to not follow the same pattern as the other (meaning, not passing a path and not using the .write method).

🤷‍♂️

cogeo_mosaic/backends/memory.py
```diff
@@ -0,0 +1,62 @@
...
+
+import attr
+from morecantile import TileMatrixSet
+from rio_tiler.constants import WEB_MERCATOR_TMS
+from rio_tiler.io import BaseReader, COGReader
+
+from cogeo_mosaic.backends.base import (
+    BaseBackend,
+    mode_validator,
+    update_zoom_and_bounds,
```
kylebarron, line 13: See this seems weird. Why not just put `mode_validator` and `update_zoom_and_bounds` on the class?

cogeo_mosaic/backends/memory.py
```diff
@@ -0,0 +1,62 @@
...
+        self.bounds = self.mosaic_def.bounds
+
+    def write(self, mosaic: MosaicJSON, overwrite: bool = True):
+        """Write mosaicjson document."""
+        if not isinstance(mosaic, MosaicJSON):
+            mosaic = MosaicJSON(**dict(mosaic))
+        self.mosaic_def = mosaic
+
+    def _read(self) -> MosaicJSON:
+        raise NotImplementedError("InMemoryBackend is a Write-Only backend.")
```
kylebarron, line 62: 😄

kylebarron (approved):

PR #171, opened 2021-04-19
Remove empty strings from a list of input files

drnextgis:
To clean up the output of `cogeo-mosaic` CLI.

```
$ cogeo-mosaic create -o mosaic.js list.txt
: No such file or directory
Get quadkey list for zoom: 10
Feed Quadkey index
```

### DIFF
--- a/cogeo_mosaic/scripts/cli.py
+++ b/cogeo_mosaic/scripts/cli.py
@@ -77,7 +77,7 @@ def create(
     quiet,
 ):
     """Create mosaic definition file."""
-    input_files = input_files.read().strip().splitlines()
+    input_files = [file.strip() for file in input_files if file.strip()]
     mosaicjson = MosaicJSON.from_urls(
         input_files,
         minzoom=minzoom,
--- a/cogeo_mosaic/scripts/cli.py
+++ b/cogeo_mosaic/scripts/cli.py
@@ -77,7 +77,7 @@ def create(
     quiet,
 ):
     """Create mosaic definition file."""
-    input_files = filter(None, input_files.read().splitlines())
+    input_files = input_files.read().strip().splitlines()
     mosaicjson = MosaicJSON.from_urls(
         input_files,
         minzoom=minzoom,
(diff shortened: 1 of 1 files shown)


cogeo_mosaic/scripts/cli.py
```diff
@@ -77,7 +77,7 @@ def create(
     quiet,
 ):
     """Create mosaic definition file."""
-    input_files = input_files.read().splitlines()
+    input_files = filter(None, input_files.read().splitlines())
```
vincentsarago, line 80: I think `input_files.read().strip()` would also work

vincentsarago, line 80:
```suggestion
    input_files = input_files.read().strip().splitlines()
```

drnextgis:
```bash
$ cat test.txt 
1.tif

2.tif
3.tif

4.tif


```

```python
$ python
>>> f = open("test.txt")
>>> f.read().strip().splitlines()
['1.tif', '', '2.tif', '3.tif', '', '4.tif']
>>> f = open("test.txt")
>>> list(filter(None, f.read().splitlines()))
['1.tif', '2.tif', '3.tif', '4.tif']
```

vincentsarago: 🤦 this will remove only empty line at the end of the file

cogeo_mosaic/scripts/cli.py
```diff
@@ -77,7 +77,7 @@ def create(
     quiet,
 ):
     """Create mosaic definition file."""
-    input_files = input_files.read().splitlines()
+    input_files = input_files.read().strip().splitlines()
```
vincentsarago, line 80:
```suggestion
    input_files = [file for file in input_files if file.strip()]
```

vincentsarago, line 80:
```suggestion
    input_files = [file.strip() for file in input_files if file.strip()]
```

vincentsarago: as @kylebarron noted (via slack) the file is opened in read mode so we don't need the `splitline` and can iterate over the lines.

We also have a preference for list comprehension

PR #184, opened 2021-10-20
add asset for bbox method

### LINKED ISSUE #181: add `assets_for_bbox` to the backend methods
in 
https://github.com/developmentseed/cogeo-mosaic/blob/0e5828e0a907cef535ff07000b0936fa91933f6e/cogeo_mosaic/backends/base.py#L119-L126

to ease integration with https://github.com/developmentseed/titiler/blob/master/src/titiler/mosaic/titiler/mosaic/factory.py#L497-L521

vincentsarago:
closes #181 

This PR also fix an issue where `get_assets` would have returned assets multiple times. We now keep asset only once but keep the order.

### DIFF
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -128,6 +128,25 @@ class BaseBackend(BaseReader):
         tile = mercantile.tile(lng, lat, self.quadkey_zoom)
         return self.get_assets(tile.x, tile.y, tile.z)
 
+    def assets_for_bbox(
+        self, xmin: float, ymin: float, xmax: float, ymax: float
+    ) -> List[str]:
+        """Retrieve assets for bbox."""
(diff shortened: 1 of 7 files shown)


cogeo_mosaic/backends/base.py
```diff
@@ -137,8 +156,10 @@ def get_assets(self, x: int, y: int, z: int) -> List[str]:
         mercator_tile = mercantile.Tile(x=x, y=y, z=z)
         quadkeys = find_quadkeys(mercator_tile, self.quadkey_zoom)
         return list(
-            itertools.chain.from_iterable(
-                [self.mosaic_def.tiles.get(qk, []) for qk in quadkeys]
+            dict.fromkeys(
```
vincentsarago, line 159: `dict.fromkeys` return `set` of data but keep the order of appearance

PR #185, opened 2021-11-18
fix `MosaicBackend` to match Backend input names and update CI

vincentsarago:


cogeo_mosaic/backends/__init__.py
```diff
@@ -14,37 +14,37 @@
 from cogeo_mosaic.backends.web import HttpBackend
 
 
-def MosaicBackend(url: str, *args: Any, **kwargs: Any) -> BaseBackend:
-    """Select mosaic backend for url."""
-    parsed = urlparse(url)
+def MosaicBackend(input: str, *args: Any, **kwargs: Any) -> BaseBackend:
```
vincentsarago, line 17: it's now called `input` in BaseBackend

PR #187, opened 2022-02-21
remove mercantile and use new morecantile

vincentsarago:
remove mercantile and use new morecantile and rio-tiler


cogeo_mosaic/backends/base.py
```diff
@@ -170,6 +170,44 @@ def get_assets(self, x: int, y: int, z: int) -> List[str]:
             )
         )
 
+    def find_quadkeys(
```
vincentsarago, line 173: moved in basebackend to use `self.tms`

PR #196, opened 2022-11-21
switch to shapely

vincentsarago:
pygeos is now part of shapely

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dynamic = ["version", "readme"]
 dependencies = [
   "attrs",
   "morecantile>=3.1,<4.0",
-  "pygeos>=0.10",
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -23,7 +23,7 @@ dynamic = ["version", "readme"]
 dependencies = [
   "attrs",
   "morecantile>=3.1,<4.0",
-  "pygeos>=0.10",
+  "shapely>=2.0b2,<2.1",
```
kylebarron, line 26: You could probably use `<3` if you trust their semver?

kylebarron (approved):
🚀

pyproject.toml
vincentsarago: well I don't trust mine 😂

PR #198, opened 2023-01-10
use az prefix for Azure Blob Storage Backend

### LINKED ISSUE #197: use `az://` prefix for azure blob storage mosaic
in https://github.com/developmentseed/titiler/discussions/569, user has an issue when trying to use public URL stored on Azure Blob Storage. In theory for public url we wouldn't need credentials but because we use `http` prefix to check if an url needs to use the ABSBackend, both public and private url will need the azure modules and credentials. 

IMO, for public URL we should always fall back to HTTPBackend and only use the ABSBackend if an user pass a url with `az://` prefix.

Note: the `az://` prefix is not `official` but that's what GDAL `/vsiaz` and Rasterio use!

https://github.com/developmentseed/cogeo-mosaic/blob/1ee9685988798a4f952fc391589dad3504548467/cogeo_mosaic/backends/__init__.py#L38-L40

cc @christoe

vincentsarago:
closes #197

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,5 @@
 
-## 5.1.0 (TBD)
+## 5.1.0 (2023-01-20)
 
 * use `az://` prefix for private Azure Blob Storage Backend.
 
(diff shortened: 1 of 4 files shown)


christoe:
Looks good to me!

PR #205, opened 2023-03-17
Update STACBackend docs

### LINKED ISSUE #204: STAC Backend documentation out of date
It looks like the `.metadata` method was [removed in version 4.0.0](https://github.com/developmentseed/cogeo-mosaic/blob/2bf5063564f685ffd64f4ce361432a1814eb659e/CHANGES.md#400a0-2021-10-20), but it is still used in the STAC Backend documentation. So that probably needs an update.

https://github.com/developmentseed/cogeo-mosaic/blob/b894f81c61864eca9843dc45a42f877638f2c999/docs/src/advanced/backends.md#stac-backend

yellowcap:
Replace `.metadata` with `.mosaic_def.dict(exclude={"tiles"})` in the print statements. Fixes #204

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
         language_version: python
 
   - repo: https://github.com/PyCQA/isort
-    rev: 5.4.2
+    rev: 5.12.0
     hooks:
       - id: isort
         language_version: python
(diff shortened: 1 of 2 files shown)


vincentsarago:
thanks @yellowcap 🙏

PR #208, opened 2023-04-04
Adds test for Mars TMS

jlaura:
This PR adds tests and support for a Mars planetary TMS. This test adds planetcantile as a dependency (to get the planetary TMS) included. That repo has a PR open [here](https://github.com/AndrewAnnex/planetcantile/pull/15) that adds the necessary `_geographic_crs` key to the TMS specifications. This ensures that the default (WGS84) that morecantile encodes is respected and that we do not get PROJ errors about missing transformations. So, [#15](https://github.com/AndrewAnnex/planetcantile/pull/15) needs to land before this lands.

Now, I am still seeing errors about incorrect negative dimensions in these tests. It looks like the bbox used in `find_extrema` is not properly populating.

Also, It is necessary to pass the TMS to the burntiles() calls in order to get the encoding working properly. So, clients are going to need to be able to pass in the TMS to cogeo-mosaic. I have not looked at if and how that happens yet. I assume that this passes in through TiTiler.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,7 @@ dynamic = ["version", "readme"]
 dependencies = [
   "attrs",
   "morecantile>=3.1,<4.0",
+  "planetcantile",
   "shapely>=2.0b2,<3.0",
   "pydantic",
   "httpx",
--- a/tests/test_planetcantile.py
+++ b/tests/test_planetcantile.py
@@ -1,0 +1,61 @@
+"""test cogeo_mosaic.supermorecado submodule.
+
+Note: most of the test are adapted from https://github.com/mapbox/supermercado/blob/main/tests/test_cli.py
+"""
+
+import planetcantile
+
+from cogeo_mosaic.supermorecado import burnTiles
+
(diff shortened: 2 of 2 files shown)


AndrewAnnex:
@jlaura I think adding planetcantile as a dependency is too heavy-handed. At most it could be considered a `dev` dependency, alternatively copying over a tms json file into the test fixtures folder for this project could be a easier solution. That said, I don't know what this test brings for cogeo-mosaic's perspective. It may make more sense to move this test into planetcantile.

jlaura:
@AndrewAnnex Sure, removing the planetcantile dependency and then loading a single TMS as a string would working fine. The larger issue with the PR is that this still is not loading the TMS properly. The round trip on the point burn is failing to convert back to projected space properly.

AndrewAnnex:
@jlaura maybe then it's worth converting the test to just use a Earth projection like EPSG:4087, may require making the TileMatrixSet object in the test if there isn't a default morecantile TMS that would work but that shouldn't be too difficult

vincentsarago:
I don't think this is relevant to cogeo-mosaic directly but will be better suited within planetcantile

PR #210, opened 2023-05-03
Start of non-WebMercatorQuad TMS support

AndrewAnnex:
* initial fixes for #202 #207 and #209
* adds cached to_rio_crs method
* updates get_footprints, get_dataset_info and tile_to_bounds to accept tms kwarg
* deprecates and removes assets_for_tile
* updates all code to use either assets_for_bbox or assets_for_point

Now, all I did so far to deprecate assets_for_tile was swap out usages directly for get_assets. Maybe the intent was to prevent all external direct-tile access? Looking for feedback on things I missed and especially on the usage of attr for the tms attribtue which I am not familiar with.

Still need to add tests with other tms specs, thinking making a custom epsg:4087 would be a good approach or any other built in morecantile tms.

PR #214, opened 2023-06-06
update for mosaicjson 0.0.3 and update morecantile

vincentsarago:
This PR takes over #210 and does:
- implement changes from the new MosaicJSON 0.0.3 specification
- allow creation of MosaicJSON for other TMS

This PR does not:
- update the Backends to use the new attributes introduces in `0.0.3`


cogeo_mosaic/backends/sqlite.py
```diff
@@ -132,7 +132,13 @@ def write(self, overwrite: bool = False):
...
                         quadkey_zoom INTEGER,
                         bounds JSON NOT NULL,
-                        center JSON
+                        center JSON,
+                        tilematrixset JSON,
+                        asset_type TEXT,
+                        asset_prefix TEXT,
+                        data_type TEXT,
+                        colormap JSON,
+                        layers JSON
```
vincentsarago, line 141: Note: This is kinda tricky but the new cogeo-mosaic version won't be able to work with sqlite mosaicJSON created with previous version. 

maybe we could add a way to `update` the metadata_table (add column)

cogeo_mosaic/mosaic.py
```diff
@@ -72,20 +72,32 @@ class MosaicJSON(BaseModel):
...
-    bounds: Tuple[float, float, float, float] = Field((-180, -90, 180, 90))
+    bounds: Tuple[float, float, float, float] = Field(default=(-180, -90, 180, 90))
     center: Optional[Tuple[float, float, int]]
     tiles: Dict[str, List[str]]
-
-    # TMS is a private attributes for now but it might change with
-    # mosaicJSON 0.3.0 (https://github.com/developmentseed/mosaicjson-spec/pull/18)
-    # Note: TMS should support quadkey (tms._is_quadtree)
-    _tms: morecantile.TileMatrixSet = PrivateAttr(default=WEB_MERCATOR_TMS)
+    tilematrixset: Optional[morecantile.TileMatrixSet]
```
vincentsarago, line 78: should we set a default?

vincentsarago: I think it's fine to assume that all mosaicJSON without tilematrixset attribute default to WebMercatorQuad 🤷

cogeo_mosaic/backends/sqlite.py
```diff
@@ -118,6 +122,11 @@ def write(self, overwrite: bool = False):
             self.delete()
 
         with self.db:
+            logger.debug(
+                f"Setting user_version to '{MOSAIC_JSON_VERSION}' in {self.db_path}."
+            )
+            self.db.execute(f"PRAGMA user_version = {MOSAIC_JSON_VERSION};")
```
vincentsarago, line 128: we can't `update` a mosaic created using the old mosaicjson specification.

In theory we could but then when we update again the specification we might regret this 😅

cogeo_mosaic/mosaic.py
```diff
@@ -171,6 +189,18 @@ def _create_mosaic(
             "version": "1.0.0",
         }
 
+        mosaic_003 = {
+            "tilematrixset": tilematrixset,
```
vincentsarago, line 193: We only add `tilematrixset` if the user provided it, if not the application should assume `WebMercatorQuad` was used

cogeo_mosaic/backends/base.py
```diff
@@ -116,6 +118,12 @@ def update(
 
         bounds = bbox_union(new_mosaic.bounds, self.mosaic_def.bounds)
 
+        if self.mosaic_def.mosaicjson != new_mosaic.mosaicjson:
+            warnings.warn(
+                f"Updating `mosaicjson` version from {self.mosaic_def.mosaicjson} to {new_mosaic.mosaicjson}"
+            )
+            self.mosaic_def.mosaicjson = new_mosaic.mosaicjson
```
vincentsarago, line 125: I think it's fine to assume that when you `update` a mosaic using this version of cogeo-mosaic, the `mosaicjson` version will be updated

cogeo_mosaic/backends/sqlite.py
```diff
@@ -192,13 +219,22 @@ def update(
         **kwargs,
     ):
         """Update existing MosaicJSON on backend."""
+        # We check the `user_version` of the Database
```
vincentsarago, line 222:
```suggestion
```

PR #215, opened 2023-07-02
add support for multiple TMS read support in Backends

vincentsarago:
continue work started in #210 

TODO
- [x] tests
- [x] documentation

cc @AndrewAnnex


cogeo_mosaic/backends/base.py
```diff
@@ -78,10 +76,17 @@ class BaseBackend(BaseReader):
     def __attrs_post_init__(self):
         """Post Init: if not passed in init, try to read from self.input."""
         self.mosaic_def = self.mosaic_def or self._read()
-        self.minzoom = self.mosaic_def.minzoom
-        self.maxzoom = self.mosaic_def.maxzoom
         self.bounds = self.mosaic_def.bounds
 
+        mosaic_tms = self.mosaic_def.tilematrixset or WEB_MERCATOR_TMS
```
vincentsarago, line 81: in order to keep support for old mosaic document we assume the default TMS to be `WebMercatorQuad`

cogeo_mosaic/backends/base.py
```diff
@@ -78,10 +76,17 @@ class BaseBackend(BaseReader):
...
+        if mosaic_tms == self.tms:
+            minzoom, maxzoom = self.mosaic_def.minzoom, self.mosaic_def.maxzoom
+        else:
+            minzoom, maxzoom = self.tms.minzoom, self.tms.maxzoom
+
```
vincentsarago, line 86: if we open the Mosaic with a TMS which is not the mosaic TMS the min/max zoom will default to the TMS (read) zooms IF min/max zoom are not passed by the user

cogeo_mosaic/backends/base.py
```diff
@@ -136,19 +142,59 @@ def update(
 
     def assets_for_tile(self, x: int, y: int, z: int) -> List[str]:
         """Retrieve assets for tile."""
-        return self.get_assets(x, y, z)
+        mosaic_tms = self.mosaic_def.tilematrixset or WEB_MERCATOR_TMS
+        if self.tms.rasterio_crs == mosaic_tms.rasterio_crs:
+            return self.get_assets(x, y, z)
```
vincentsarago, line 147: if the Read TMS is the same as the Mosaic TMS we can directly use `get_assets`

cogeo_mosaic/backends/base.py
```diff
@@ -136,19 +142,59 @@ def update(
...
+
+        xmin, ymin, xmax, ymax = self.tms.bounds(x, y, z)
+        return self.assets_for_bbox(
+            xmin,
+            ymin,
+            xmax,
+            ymax,
+            coord_crs=self.tms.rasterio_geographic_crs,
+        )
```
vincentsarago, line 156: if Read TMS is different we then use `assets_for_bbox`

cogeo_mosaic/backends/base.py
```diff
@@ -136,19 +142,59 @@ def update(
 
     def assets_for_tile(self, x: int, y: int, z: int) -> List[str]:
         """Retrieve assets for tile."""
-        return self.get_assets(x, y, z)
+        mosaic_tms = self.mosaic_def.tilematrixset or WEB_MERCATOR_TMS
+        if self.tms.rasterio_crs == mosaic_tms.rasterio_crs:
+            return self.get_assets(x, y, z)
+
+        xmin, ymin, xmax, ymax = self.tms.bounds(x, y, z)
```
vincentsarago, line 149: use geographic coordinates or TMS's crs ?

cogeo_mosaic/backends/base.py
```diff
@@ -136,19 +142,59 @@ def update(
...
+        mosaic_tms = self.mosaic_def.tilematrixset or WEB_MERCATOR_TMS
+        if coord_crs != mosaic_tms.rasterio_geographic_crs:
+            xmin, ymin, xmax, ymax = transform_bounds(
+                coord_crs,
+                mosaic_tms.rasterio_geographic_crs,
+                xmin,
+                ymin,
+                xmax,
+                ymax,
+            )
```
vincentsarago, line 194: transform the coordinates is Mosaic's geographic coordinates 

TBD: use Mosaic's CRS?

vincentsarago:
@AndrewAnnex this is ready for review if you want to try 🙏 

(FYI: I will need to merge this before the end of the week)

cogeo_mosaic/backends/base.py
```diff
@@ -38,50 +38,60 @@ class BaseBackend(BaseReader):
...
+        maxzoom (int): mosaic Max zoom level. Defaults to tms or mosaic maxzoom.
         reader (rio_tiler.io.BaseReader): Dataset reader. Defaults to `rio_tiler.io.Reader`.
         reader_options (dict): Options to forward to the reader config.
         geographic_crs (rasterio.crs.CRS, optional): CRS to use as geographic coordinate system. Defaults to WGS84.
-        tms (morecantile.TileMatrixSet, optional): TileMatrixSet grid definition. **READ ONLY attribute**. Defaults to `WebMercatorQuad`.
-        bbox (tuple): mosaic bounds (left, bottom, right, top). **READ ONLY attribute**. Defaults to `(-180, -90, 180, 90)`.
-        minzoom (int): mosaic Min zoom level. **READ ONLY attribute**. Defaults to `0`.
-        maxzoom (int): mosaic Max zoom level. **READ ONLY attribute**. Defaults to `30`
+        bounds (tuple): mosaic bounds (left, bottom, right, top). **READ ONLY attribute**. Defaults to `(-180, -90, 180, 90)`.
+        crs (rasterio.crs.CRS): mosaic crs (left, bottom, right, top). **READ ONLY attribute**. Defaults to `(-180, -90, 180, 90)`.
```
AndrewAnnex, line 48: defaults comment here seems to be copied from the line above

cogeo_mosaic/backends/base.py
```diff
@@ -38,50 +38,60 @@ class BaseBackend(BaseReader):
...
-    tms: TileMatrixSet = attr.ib(init=False, default=WEB_MERCATOR_TMS)
-    minzoom: int = attr.ib(init=False)
-    maxzoom: int = attr.ib(init=False)
-
-    # default values for bounds
     bounds: Tuple[float, float, float, float] = attr.ib(
         init=False, default=(-180, -90, 180, 90)
     )
     crs: CRS = attr.ib(init=False, default=WGS84_CRS)
+    geographic_crs: CRS = attr.ib(init=False, default=WGS84_CRS)
```
AndrewAnnex, line 70: there is a number of places where the crs and geographic_crs are defined, but shouldn't the geographic_crs always correspond to the one used by the crs (projection or otherwise)? So it would be easier to just only ever explicitly define the crs and always do `crs.geographic_crs` unless I am missing something?

cogeo_mosaic/backends/base.py
AndrewAnnex: a number of comments made on the pr would be better as inline comments in the code

cogeo_mosaic/backends/base.py
```diff
@@ -38,50 +38,60 @@ class BaseBackend(BaseReader):
...
+        # Geographic CRS.
+        self.crs = mosaic_tms.rasterio_geographic_crs
+        self.geographic_crs = mosaic_tms.rasterio_geographic_crs
+
+        if mosaic_tms == self.tms:
+            minzoom, maxzoom = self.mosaic_def.minzoom, self.mosaic_def.maxzoom
+        else:
+            minzoom, maxzoom = self.tms.minzoom, self.tms.maxzoom
+
+        self.minzoom = self.minzoom if self.minzoom is not None else minzoom
```
AndrewAnnex, line 92: easier to do `if self.minzoom is None: \n self.minzoom = minzoom` maybe?

cogeo_mosaic/backends/base.py
AndrewAnnex: was a bit confused by this from prior prs, if tms.bounds returns coordinates in the crs of the tms then it should be fine as is. Using one or the other shouldn't matter so long as they are used with consistency

cogeo_mosaic/backends/base.py
```diff
@@ -136,19 +147,59 @@ def update(
...
+    def assets_for_point(
+        self,
+        lng: float,
+        lat: float,
+        coord_crs: CRS = WGS84_CRS,
+    ) -> List[str]:
         """Retrieve assets for point."""
-        tile = self.tms.tile(lng, lat, self.quadkey_zoom)
+        mosaic_tms = self.mosaic_def.tilematrixset or WEB_MERCATOR_TMS
+        if coord_crs != mosaic_tms.rasterio_geographic_crs:
```
AndrewAnnex, line 171: aren't all long/lats always geographic coordinates? Maybe the check here could allow for projected coords also (northing/easting/etc).

cogeo_mosaic/backends/memory.py
```diff
@@ -38,10 +47,23 @@ class MemoryBackend(BaseBackend):
 
     def __attrs_post_init__(self):
         """Post Init."""
-        self.minzoom = self.mosaic_def.minzoom
-        self.maxzoom = self.mosaic_def.maxzoom
         self.bounds = self.mosaic_def.bounds
 
+        mosaic_tms = self.mosaic_def.tilematrixset or WEB_MERCATOR_TMS
```
AndrewAnnex, line 52: a lot of dupe code here from the base class, any way to combine?

cogeo_mosaic/backends/stac.py
```diff
@@ -84,6 +91,13 @@ def __attrs_post_init__(self):
         self.mosaic_def = self._read()
         self.bounds = self.mosaic_def.bounds
 
+        mosaic_tms = self.mosaic_def.tilematrixset or WEB_MERCATOR_TMS
+
+        # By mosaic definition the bounds and CRS are defined using the TMS
+        # Geographic CRS.
+        self.crs = mosaic_tms.rasterio_geographic_crs
```
AndrewAnnex, line 98: think this is a typo, as both crss are the geographic_crs here

AndrewAnnex (commented):
only actually a partial review due to other plans today but left some questions and suggestions to get started

cogeo_mosaic/backends/base.py
vincentsarago: > but shouldn't the geographic_crs always correspond to the one used by the crs (projection or otherwise)? 

Sadly in Rasterio's CRS the `geodetic_crs` property is not available which is why we have separate property in rio-tiler (https://github.com/cogeotiff/rio-tiler/blob/main/rio_tiler/io/base.py#L42-L44)

cogeo_mosaic/backends/stac.py
vincentsarago: > as both crss are the geographic_crs here

Yes. 
By definition the `bounds` property defined in the mosaic document use WGS84 (or geographic CRS) while in rio-tiler when using any dataset the `bounds` is using the dataset's `crs` which is why we needed a `geographic_crs` property 

FYI: this is used there https://github.com/cogeotiff/rio-tiler/blob/6d71e7d380cd815b8f65a1917653769d882c2aa6/rio_tiler/io/base.py#L90-L108

vincentsarago: I understand this is confusing but it comes from the mosaicjson definition 😓

cogeo_mosaic/backends/base.py
vincentsarago: good question. 

in rio-tiler we always assume the coordinate to be geographic and in X, Y order 🤷. 

my understanding is that user should always provide coordinates in X, Y order

cogeo_mosaic/backends/base.py
AndrewAnnex: ahh forgot it wasn't a Proj CRS

cogeo_mosaic/backends/base.py
AndrewAnnex: okay forgot that we want to allow for other coord_crss here so lgtm

cogeo_mosaic/backends/base.py
AndrewAnnex: maybe I am forgetting the difference between a rasterio and pyproj crs here, but at least with proj it should be safe to directly convert without first going to geographic coords here

vincentsarago: The question was more, should we first convert to CRS ([`tms._tile`](https://github.com/developmentseed/morecantile/blob/ecabdd1f4a22a8f85e4d8721d03e20d621e64ce6/morecantile/models.py#L717)) instead of Geographic CRS ([`tms.tile`](https://github.com/developmentseed/morecantile/blob/ecabdd1f4a22a8f85e4d8721d03e20d621e64ce6/morecantile/models.py#L769)) later in the code.

Going back and forth from geographic to local can be tricky so I think it will be safer to transform the coordinates to  the tms's CRS (instead of Geographic CRS) and then use `tms._tile` method

cogeo_mosaic/backends/memory.py
```diff
@@ -22,26 +23,28 @@ class MemoryBackend(BaseBackend):
 
     mosaic_def: MosaicJSON = attr.ib(converter=_convert_to_mosaicjson)
 
+    tms: TileMatrixSet = attr.ib(default=WEB_MERCATOR_TMS)
```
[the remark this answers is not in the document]
vincentsarago: because there are some difference between input attribute with/without default we do need to re-write most of the attribute 😓

cogeo_mosaic/backends/base.py
vincentsarago: edited ☝️ 

I think the current implementation is fine and is what we had earlier.

vincentsarago:
alright thanks for the review @AndrewAnnex 🙏 
Let me know if you want to have another look or maybe try yourself before we merge this

AndrewAnnex:
@vincentsarago nope, I think it's good to go, or rather, I don't have time to test it prior to the deadline you mentioned. I'm sure any fixes that are needed would be small

PR #219, opened 2023-07-11
fix post init for HTTPBackend

vincentsarago:


cogeo_mosaic/backends/web.py
```diff
@@ -25,17 +25,10 @@ class HttpBackend(BaseBackend):
 
     # Because the HttpBackend is a Read-Only backend, there is no need for
     # mosaic_def to be in the init method.
-    mosaic_def: MosaicJSON = attr.ib(init=False)
+    mosaic_def: MosaicJSON = attr.ib(init=False, default=None)
```
vincentsarago, line 28: with `mosaic_def` set to None, the default post_init method from the BaseBackend will do all the work

PR #233, opened 2024-09-24
Add tms parameter to cli for MosaicJSON

AndrewAnnex:
I've added the TMS as an optional string parameter (either a path to json or an id morecantile can look up) to most of the mosaic json cli functions. This also now changes the default tms to always be specified to at least WebMercatorQuad  although I haven't worked out the best option for serialization of the the tms in the mosaic json output, but likely dumping the full tms to json is best. Likely a lot of corner cases are present here I haven't thought of so consider this a WIP and I would love help getting this in.


cogeo_mosaic/mosaic.py
```diff
@@ -275,7 +275,13 @@ def from_urls(
             >>> MosaicJSON.from_urls(["1.tif", "2.tif"])
 
         """
-        features = get_footprints(urls, max_threads=max_threads, quiet=quiet)
+        tms = kwargs.pop('tilematrixset', WEB_MERCATOR_TMS)
```
vincentsarago, line 278: can we add the `tms` option in the function definition

same as in rio-cogeo https://github.com/cogeotiff/rio-cogeo/blob/main/rio_cogeo/cogeo.py#L86
https://github.com/cogeotiff/rio-cogeo/blob/main/rio_cogeo/cogeo.py#L194

AndrewAnnex: went with `tilematrixset` as the parameter name, otherwise other usages like in test_mosaic_create_tms are not valid and would require updates, so hopefully less of a breaking change this way. That said the inconsistency between some functions expecting tms and other expecting tilematrixset is problematic

AndrewAnnex: @vincentsarago do you think it's worth renaming the parameter to `tms` here and elsewhere, the inconsistency is something I'd like to correct, but have no strong leaning which name to pick other than a preference for being explicit

cogeo_mosaic/mosaic.py
```diff
@@ -271,7 +273,9 @@ def from_urls(
             >>> MosaicJSON.from_urls(["1.tif", "2.tif"])
 
         """
-        features = get_footprints(urls, max_threads=max_threads, quiet=quiet)
+        features = get_footprints(
+            urls, max_threads=max_threads, tms=tilematrixset, quiet=quiet
+        )
```
vincentsarago, line 278: tms will default to `WebMercatorQuad` directly in the `get_footprints` method

cogeo_mosaic/mosaic.py
```diff
@@ -297,7 +301,12 @@ def from_urls(
             raise MultipleDataTypeError("Dataset should have the same data type")
 
         return cls._create_mosaic(
-            features, minzoom=minzoom, maxzoom=maxzoom, quiet=quiet, **kwargs
+            features,
+            minzoom=minzoom,
+            maxzoom=maxzoom,
+            tilematrixset=tilematrixset,
```
vincentsarago, line 307: if None, tms will default to `WebMercatorQuad` directly in the `_create_mosaic` method

vincentsarago:
The tests fails because now the CLI will add tilematrixset into the MosaicJSON (while before we were not setting it in the mosaicJSON, assuming None meant WebMercatorQuad)

I need to think a bit more about this

cogeo_mosaic/scripts/cli.py
```diff
@@ -92,6 +108,7 @@ def create(
         quadkey_zoom=quadkey_zoom,
         minimum_tile_cover=min_tile_cover,
         tile_cover_sort=tile_cover_sort,
+        tilematrixset=tilematrixset,
```
vincentsarago, line 111: the default will be `None` and will only be set when passing `--tms ...` 

internally, cogeo-mosaic will default to WebMercatorQuad but without setting it in the MosaicJSON file

vincentsarago:
@AndrewAnnex I've made some change to keep it compatible with older version. 

It would be nice to have some tests with non-Earth TMS to validate things 🙏

cogeo_mosaic/mosaic.py
AndrewAnnex: not sure what you wanted changed here and in related comments. Do you want to add an inline comment?

AndrewAnnex:
> @AndrewAnnex I've made some change to keep it compatible with older version.
> 
> It would be nice to have some tests with non-Earth TMS to validate things 🙏

will do

vincentsarago (approved):

PR #234, opened 2024-09-29
update BaseBackend to use a default coord_crs from the tms 

AndrewAnnex:
update BaseBackend to use a default coord_crs from the tms parameter. There is also the mosaic def's tms which could be used but it is not clear if should be assumed or not...

This was causing issues with custom non earth TMSs. This is a higher priority than #233 to get resolved.

### DIFF
--- a/cogeo_mosaic/backends/base.py
+++ b/cogeo_mosaic/backends/base.py
@@ -178,8 +178,7 @@ class BaseBackend(BaseReader):
         """Retrieve assets for point."""
         mosaic_tms = self.mosaic_def.tilematrixset or WEB_MERCATOR_TMS
         # default coord_sys should be the mosaic's geographic CRS
-        if not coord_crs:
-            coord_crs = self.tms.rasterio_geographic_crs
+        coord_crs = coord_crs or self.tms.rasterio_geographic_crs
         # If coord_crs is not the same as the mosaic's geographic CRS
         # we reproject the coordinates
         if coord_crs != mosaic_tms.rasterio_geographic_crs:
@@ -204,8 +203,7 @@ class BaseBackend(BaseReader):
         """Retrieve assets for bbox."""
         mosaic_tms = self.mosaic_def.tilematrixset or WEB_MERCATOR_TMS
         # default coord_sys should be the mosaic's geographic CRS
-        if not coord_crs:
-            coord_crs = self.tms.rasterio_geographic_crs
+        coord_crs = coord_crs or self.tms.rasterio_geographic_crs
         # If coord_crs is not the same as the mosaic's geographic CRS
         # we reproject the bounding box
         if coord_crs != mosaic_tms.rasterio_geographic_crs:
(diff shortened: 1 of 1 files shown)


cogeo_mosaic/backends/base.py
```diff
@@ -173,11 +173,13 @@ def assets_for_point(
...
         lat: float,
-        coord_crs: CRS = WGS84_CRS,
+        coord_crs: Optional[CRS] = None,
     ) -> List[str]:
         """Retrieve assets for point."""
         mosaic_tms = self.mosaic_def.tilematrixset or WEB_MERCATOR_TMS
-
+        # default coord_sys should be the mosaic's geographic CRS
+        if not coord_crs:
+            coord_crs = self.tms.rasterio_geographic_crs
```
vincentsarago, line 182: we usually do `coord_crs = coord_crs or self.tms.rasterio_geographic_crs` (it saves one line 😅 )

AndrewAnnex:
@vincentsarago, forgive me if I don't explain the issue well here as it's fairly complicated (lots of interplay between Proj/WKT2 standards) but here's my current understanding:

I am working in the planetcantile project to define TMSs for other planetary bodies (Moon, Mars, etc) so that these can be used across titiler/morecantile/cogeo-mosaic for mosaic.json. Currently in Titiler the mosaicjson endpoints will default to WGS84 for the CRS here and cause Proj errors when attempting to convert between my non-earth CRSs/TMSs.

 o without modifying Titiler directly I think it makes more sense to update cogeo-mosaic as each TMS has a explicit or implicitly discoverable geographic crs from it's CRS parameter. One can work around this by setting `PROJ_IGNORE_CELESTIAL_BODY=YES`, but that is a hack and may not always do the right thing depending on the ellipsoid (many of the planetary CRSs have ellipsoids with different flattenings/ no fattenings as the ellipsoids are defined with equal semi major/minor axes, but are not explicitly spherical CRS which are possible to define with WKT2...)

Again, If a person defines a mosaic.json file, a TMS is required, and so think it makes sense to always grab the appropriate geographic crs from the TMS object (which will know it implicitly from the CRS/geographic_crs of the tms) rather than assuming WGS84. `assets_for_tile` seems to do this already but for consistency and to avoid the exceptions I was seeing from titiler this PR attempts to make that consistent

On a side note, I am also debugging an issue in morecantile with the geographic_crs/rasterio_geographic_crs ending up as 4326 vs the intended geographic_crs depending on how the TMS is serialized/initialized. Although I haven't made enough progress on that to post an issue or pr yet.

vincentsarago (approved):

cogeo_mosaic/backends/base.py
```diff
@@ -173,11 +173,12 @@ def assets_for_point(
         self,
         lng: float,
         lat: float,
-        coord_crs: CRS = WGS84_CRS,
+        coord_crs: Optional[CRS] = None,
     ) -> List[str]:
         """Retrieve assets for point."""
         mosaic_tms = self.mosaic_def.tilematrixset or WEB_MERCATOR_TMS
-
+        # default coord_sys should be the mosaic's geographic CRS
```
vincentsarago, line 180:
```suggestion
        # default coord_crs should be the TMS's geographic CRS
```

cogeo_mosaic/backends/base.py
```diff
@@ -197,11 +198,12 @@ def assets_for_bbox(
         ymin: float,
         xmax: float,
         ymax: float,
-        coord_crs: Optional[CRS] = WGS84_CRS,
+        coord_crs: Optional[CRS] = None,
     ) -> List[str]:
         """Retrieve assets for bbox."""
         mosaic_tms = self.mosaic_def.tilematrixset or WEB_MERCATOR_TMS
-
+        # default coord_sys should be the mosaic's geographic CRS
```
vincentsarago, line 205:
```suggestion
        # default coord_crs should be the TMS's geographic CRS
```

